PR #1, opened 2016-08-06
Enable frozen_string_literal

mizoR:
Fix a following warning and enable frozen_string_literal:

```sh
$ find . | grep -e '\.rb$' | xargs ruby -w
./lib/injector.rb:25: warning: `frozen_string_literal' is ignored after any tokens
```


wyukawa:
Merged.
Thanks.

PR #7, opened 2016-08-25
Add a page to show a single service with it's list of projects

kfdm:
Since the main / page can get very long after many service and projects have been added, it's useful to have a page to only view a single service by itself

### DIFF
--- a/views/show_service.erb
+++ b/views/show_service.erb
@@ -1,0 +1,45 @@
+<div class="page-header">
+  <h1>Service - <%= @service.name %></h1>
+  <div class="btn-group" role="group">
+    <a href="/service/register" class="btn btn-default">Register New Service</a>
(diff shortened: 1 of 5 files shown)


tokuhirom:
👍

views/show_service.erb
```diff
@@ -0,0 +1,45 @@
...
+            <th><%= exporter.job %></th>
+            <td><%= exporter.port %></td>
+          </tr>
+          <% end %>
+        </table>
+      </td>
+      <td><%= @project_id2farm_name[project.id] %></td>
+    </tr>
+    <% end %>
+  </table>
```
mizoR: lack of `</div>`

kfdm: Ah, nice catch. Thank you.

PR #9, opened 2016-09-13
Metrics stub to allow Prometheus to check up/down status

kfdm:
Later I would like to implement some promgen specific metrics (particularly on Alert::Sender delivery) but in the mean time, I thought it would be good to have an empty /metrics route so that I could take advantage of Prometheus and a basic InstanceDown rule to ensure that Promgen is running

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -13,6 +13,8 @@ app = Promgen.build
 use Rack::Static, urls: ['/css'], root: File.join(File.dirname(__FILE__), 'public')
 
 run Rack::URLMap.new(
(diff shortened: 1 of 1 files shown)


wyukawa:
Could you please add test?

kfdm:
Since this stub is handled at the Rack level instead of the Sinatra layer, it's difficult to write a test case for this. Refactoring to a Sinatra class just to provide a test adds a lot more code to maintain. It seems a little strange to refactor and add ~50 lines of code to maintain, just to test a small stub endpoint :)

wyukawa:
Ah, I understand.

PR #10, opened 2016-09-13
Audit log page

kfdm:
There are probably various style bits that will need to be fixed, but this adds a new audit log page to see what actions have been changed recently.

One place that I need feedback on, is the specific content of some of the logging messages, though some of that can be handled in a different pull request

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'mysql', require: false
 gem 'chronic_duration'
 gem 'mail'
 gem 'unicorn'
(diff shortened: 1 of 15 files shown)


wyukawa:
Overall, it seems good.

But "Registered exporter node:9100 to test:test" is a little confused in my feeling.

It's better to clarify what is service/project.
For example, "Registered exporter node:9100 to service test and project test"

And could you please add test?

tokuhirom:
awesome!

PR #13, opened 2016-09-16
Make metrics path configurable

kojilin:
Apply this change will make metrics path configurable, this will output json file like
```
{
  "targets":[
    "localhost:5000"
  ],
  "labels":{
    "service":"foo",
    "project":"bar",
    "farm":"hoge",
    "job":"hogehoge",
    "__metrics_path__":"/internal/metrics"
  }
}
```

### DIFF
--- a/lib/promgen/web/route/project_exporter_route.rb
+++ b/lib/promgen/web/route/project_exporter_route.rb
@@ -33,12 +33,13 @@ class Promgen
     post '/project/:project_id/exporter/register' do
       port = params['port']
       job = params['job']
+      path = empty_to_nil(params[:path])
 
-      @project_exporter_service.register(project_id: @project.id, port: port, job: job)
+      @project_exporter_service.register(project_id: @project.id, port: port, job: job, path: path)
 
       @project = @project_service.find(id: @project.id)
       @service = @service_service.find(id: @project.service_id)
-      @audit_log_service.log(entry: "Registered exporter #{job}:#{port} to Service:#{@service.name}/Project:#{@project.name}")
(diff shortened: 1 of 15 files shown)


views/show_project.erb
```diff
@@ -48,14 +48,17 @@
...
           <th>&nbsp;</th>
         </tr>
         <% @exporters.each do |exporter| %>
             <tr>
               <td><%= exporter.job %></td>
               <td><%= exporter.port %></td>
+              <td><%= exporter.path %></td>
               <td>
                 <form method="post" action="<%= to("/project/#{@project.id}/exporter/#{exporter.port}/delete") %>" onsubmit="return confirm('Delete this exporter?')">
+                  <input type="hidden"i name="path" value="<%= exporter.path %>"/>
```
kfdm: Looks like there's an extra "i" here.

lib/promgen/web/route/project_exporter_route.rb
```diff
@@ -34,7 +34,7 @@ class Web < Sinatra::Base
       port = params['port']
       job = params['job']
 
-      @project_exporter_service.register(project_id: @project.id, port: port, job: job)
+      @project_exporter_service.register(project_id: @project.id, port: port, job: job, path: empty_to_nil(params[:path]))
 
       @project = @project_service.find(id: @project.id)
       @service = @service_service.find(id: @project.service_id)
```
kfdm: Can you add mention of the metrics path to the ```@audit_log_service``` entry

views/show_project.erb
kojilin: ....oops vim..

wyukawa:
Could you please add screen shot?

kojilin:
updated, and I'm not familiar with sequel. I don't know how to drop unique constraint elegantly, so just modified migration/version 5 to let unique can have a name.

PR #18, opened 2016-09-28
Trigger write api

kfdm:
Since Promgen manages the configuration file for Prometheus, it’s normal for Promgen and Prometheus to be paired together on the same server. Once you have multiple servers for load balancing and failover, it becomes more difficult to keep the configuration files in sync.

This pull request adds two enhancements.

1) POSTing to /api/v1/config and /api/v1/rule/ will trigger to Promgen to write it’s configuration changes to disk.

2) Add notification hooks to config_writer and rule_writer to POST updates to configured endpoints on updates.

```yaml
# Config for promgen1.local
config_writer:
  path: /tmp/prom.json
  notify:
    - http://promgen.other/api/v1/config
rule_writer:
  rule_path: /tmp/prom.rule
  promtool_path: /path/to/promtool
  notify:
    - http://promgen.other/api/v1/rule/
```

Assuming that each Promgen instance shares the same database, this allows one to have multiple, load balanced promgen instances where trigging a write on one instance, will also trigger writes on the other instances.

I also moved the API routes to be it’s own controller (instead of a child of app.web) to allow different Rack settings to be applied separate from /alerts and other web endpoints.

Since the Promgen POST endpoints ignore the POST body, it would be easy to add a Pushgateway compatible body, and have Promgen send a last_updated_timestamp to Pushgateway to be used in other metrics or notifications

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,26 @@ When you specify the Webhook module, you need to specify the URL for each projec
 
 Specify the Alerta module and the URL set in your environment.
 
+### 6. Notification Hooks
+
+When Promgen writes it's target configuration file or alerting rules file, it can optionally post a notification to an external service.
+This is primarily useful for keeping configuration files in sync while running multiple Prometheus+Promgen servers in parallel
+
+```yaml
+config_writer:
+  path: /tmp/prom.json
+  notify:
+    - http://promgen.other/api/v1/config
+rule_writer:
+  rule_path: /tmp/prom.rule
+  promtool_path: /path/to/promtool
+  notify:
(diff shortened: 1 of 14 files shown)


wyukawa:
Overall, it seems good.

To add document would be nice.

Could you please add Trigger write api document to README and API Docs(api.erb)?

Although too much document is not necessary,  for example, to add explanation of ```notify``` to README is better.

kfdm:
Updated the README and /api/ page to hopefully explain things a bit better :)

wyukawa:
Did you push?

PR #19, opened 2016-09-30
Add LINE Notify notification

wyukawa:
## Motivation

LINE Notify has been released.
https://notify-bot.line.me

So we can receive alert notifications through LINE Notify.

This pull request is the implementation.

## Design

Access token is stored in DB and is not displayed in Web UI because of safety.

curl example
```
$ curl -X POST -H 'Authorization: Bearer <access_token>' -F 'message=foobar' https://notify-api.line.me/api/notify
{"status":200,"message":"ok"}
```

## Screen Shots

create project

list service

show project

LINE Notify


views/list_project.erb
```diff
@@ -16,13 +16,15 @@
...
       <th>Webhook URL</th>
+      <th>Access Token</th>
     </tr>
     <% @projects.each do |project| %>
       <tr>
         <td><a href="<%= to("/project/#{ project.id }") %>"><%= project.name %></a></td>
         <td><%= project.hipchat_channel %></td>
         <td><%= project.mail_address %></td>
         <td><%= project.webhook_url %></td>
+        <td>***</td>
```
kfdm: This should probably be an if statement

```ruby
{% if project. access_token%}***{%endif %}
```

Otherwise it's confusing if this is set or not

kfdm: Same comment for the other places this shows up. Should be an easy fix though :)

views/update_project.erb
```diff
@@ -26,6 +26,10 @@
       <label for="webhookURL">Webhook URL</label>
       <input type="text" name="webhook_url" class="form-control"  id="webhookURL" value="<%= @project.webhook_url %>">
     </div>
+    <div class="form-group">
+      <label for="inputAccessToken">Access Token</label>
+      <input type="password" name="access_token" class="form-control"  id="inputAccessToken" value="<%= @project.access_token %>">
```
be-hase: `value="<%= @project.access_token %>"`

hmm...accessToken is exposed on source html.

wyukawa:
Thank you for comments

I modify to use the following code
```
<% if project.line_notify_access_token %>***<% end %>
```

screen shots

views/update_project.erb
```diff
@@ -26,6 +26,10 @@
       <label for="webhookURL">Webhook URL</label>
       <input type="text" name="webhook_url" class="form-control"  id="webhookURL" value="<%= @project.webhook_url %>">
     </div>
+    <div class="form-group">
+      <label for="inputLineNotifyAccessToken">LINE Notify Access Token</label>
+      <input type="password" name="line_notify_access_token" class="form-control"  id="inputLineNotifyAccessToken" value="<% if @project.line_notify_access_token %>***<% end %>">
```
kfdm: This part is a little trickier if you want to show that it is filled in, without overwriting the token with "***" each time they save.

For starters, you could try to use the placeholder attribute ```<% if @project.line_notify_access_token %>placeholder="***"><% end %>``` to show that it is set or not. You will also likely need a button or checkbox to detect between when you want to clear the password and when you want to update the password. I will try to write up a quick example in a bit.

wyukawa: Thank you for your comments!

Actually, I understand that current implementation is bad because for example, if user try to change a hipchat room without editing LINE Notify Access Token, user involuntarily sets the access token to '***'.

So I think that it's better to divide '/project/*/update' screen.

* show_project


* update_project


* update_line_notify_access_token


Even though this implementation, if user push the Update button without editing token, '***' is set.
But I hope no such users...

What do you think?

views/update_line_notify_access_token.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<div class="page-header">
+  <h1>Update project configuration - <%= @project.name %></h1>
+</div>
+
+<div class="row">
+  <form method="post" action="<%= to("/project/#{@project.id}/update_line_notify_access_token") %>">
+    <div class="form-group">
+      <label for="inputLineNotifyAccessToken">LINE Notify Access Token</label>
+      <input type="password" name="line_notify_access_token" class="form-control"  id="inputLineNotifyAccessToken" value="<% if @project.line_notify_access_token %>***<% end %>">
```
kfdm: I think things look good, separating things out I think is the right idea, since it makes things a bit easier to understand. You should remove the value="" part here, and probably assume if a user makes it to this page, they either want to clear the token (submit an empty value) or submit a new one. Adding a default value here means that if a user navigates here and submits, then they would set their token to "***" which is probably not what they intended to do.

```
<input type="password" name="line_notify_access_token" class="form-control"  id="inputLineNotifyAccessToken">

```

should be sufficient

wyukawa: Thank you for your inspection.
I agree with you.
I fix.

PR #37, opened 2018-01-26
Dockerfile: Ship promtool in docker image

seoester:
I noticed that the docker image doesn't contain `promtool`, even though it is required for rule checking by the web app.

I extended the Dockerfile to download `promtool` from the current Prometheus release binaries (that is [v2.1.0](https://github.com/prometheus/prometheus/releases)) and place it in `/usr/local/bin/promtool`.

Note: This PR contains the Prometheus version hard-coded in the Dockerfile. Alternatively, the latest version could be fetched using the Github API.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,19 @@
 FROM python:3.5.3-alpine
 MAINTAINER paul.traylor@linecorp.com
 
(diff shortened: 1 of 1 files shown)


seoester:
Closed and reopened as #38.

PR #38, opened 2018-01-27
Dockerfile: Ship promtool in docker image

seoester:
I noticed that the docker image doesn't contain `promtool`, even though it is required for rule checking by the web app.

I extended the Dockerfile to download `promtool` from the current Prometheus release binaries (that is [v2.1.0](https://github.com/prometheus/prometheus/releases)) and place it in `/usr/local/bin/promtool`.

Note: This PR contains the Prometheus version hard-coded in the Dockerfile. Alternatively, the latest version could be fetched using the Github API.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,19 @@
 FROM python:3.5.3-alpine
 MAINTAINER paul.traylor@linecorp.com
 
+ENV PROMETHEUS_VERSION 2.1.0
+ENV PROMETHEUS_DOWNLOAD_URL https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz
+
 RUN adduser -D -u 1000 promgen promgen
 RUN apk add --no-cache --update mariadb-dev build-base bash && \
     rm -rf /var/cache/apk/*
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,20 @@
 FROM python:3.5.3-alpine
 MAINTAINER paul.traylor@linecorp.com
 
+ENV PROMETHEUS_VERSION 2.1.0
```
kfdm: > Note: This PR contains the Prometheus version hard-coded in the Dockerfile. Alternatively, the latest version could be fetched using the Github API.

I think hardcoding this is not a problem. Ultimately I added this to help check syntax errors and such before deploying to Prometheus so it likely does not always need to be the absolute latest version.

kfdm:
Thank you for this change!
Internally I still deploy directly (into a virtualenv instead of docker) but this should be simplify things for those who have already embraced containers :)

seoester:
Thank you for merging my requests so quickly.

The docker setup is important for me, as we are considering deploying promgen for managing our primary prometheus instance. It strikes a good balance between automation and manual configuration, thanks a lot for your work :)

Unfortunately, the docker instructions in both the README and the sphinx doc are not complete. I wrote a new docker-compose file which I will PR here in the next days. Hopefully, it can help others with trying out promgen more easily.

PR #39, opened 2018-01-27
Fixes related to rule writing

seoester:
The changes allow the writing of the rule file by the promgen-worker using the default configuration.


kfdm:
Good catch on fixing things here. I had fixed a few things in b89715674e534e86e3017119e6ba80f50a57b202 but now this reminds me I need to fix my test case for the v1 + v2 formats so I will try to do that tomorrow. Thanks again for the PR :)

kfdm:
The other PR should be merged so I'll close this, but thanks again for catching this! :)

PR #40, opened 2018-01-28
Added slack notifier

seoester:
Added slack notifier.

The notifier requires a [Slack Webhook](https://api.slack.com/incoming-webhooks) to be configured in advance.

### DIFF
--- a/promgen/notification/slack.py
+++ b/promgen/notification/slack.py
@@ -1,0 +1,50 @@
+# These sources are released under the terms of the MIT license: see LICENSE
+
+import logging
+
+from django import forms
+
+from promgen import util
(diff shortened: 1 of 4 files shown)


seoester:
Happy to share this integration, I imagine many people may want to integrate with Slack these days.

I only glanced at the test cases so far. I'll check if I can make sense of them and write a test case for the slack one.

Regarding the documentation: I think it might just be slightly out of date. I ended up simply copying the linenotify file and chaging the relevant bits. The `Form…` could use some explanation, as well as how it relates to the first parameter passed to `_send`. Also: I think all senders / notifiers are executed in-line and don't use celery to be run by a worker?

PR #42, opened 2018-01-29
Django conf docs

seoester:
One more pull request for your consideration ;)

This one adds a new page to the docs about configuring django settings. I added an example how to configure SMTP server settings.

### DIFF
--- a/docs/django-conf.rst
+++ b/docs/django-conf.rst
@@ -1,0 +1,35 @@
+Configuring Django
+==================
+
+Arbitrary `django <https://www.djangoproject.com/>`__ settings can be
+set for the promgen web app by adding those under the ``django`` key to
+the ``promgen.yml`` file.
+
+All available django settings (not every setting may apply to promgen)
+are listed in the `django
+reference <https://docs.djangoproject.com/en/2.0/ref/settings/>`__.
+
+Configuring an SMTP Server
+--------------------------
+
+An SMTP server for sending outgoing mail can be configured this way:
+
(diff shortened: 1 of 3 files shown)


kfdm:
Thanks for the docs updates! It's always difficult to remember all the steps required to setup from scratch ^^;

I originally had put several settings (like database) in ENV variables ( ~/.config/promgen ) to try to match a [12 factor app](https://12factor.net/config) though with some of the settings a config file is actually easier so perhaps I should try to update the documentation to suggest that as the recommended way. Do you have any thoughts on ENV vars vs configuration file?

seoester:
Sorry, I just rediscovered the question at the end of your comment.

The 12 factor explanation is simple enough, however, I'd like to make an argument for conf files.

For many applications configuration doesn't just consist of information about the immediate run-time environment but also some "business" data or logic. Furthermore, some configuration options are very specific and deeply integrated into the application.

Env vars are rather generic as they are language-agnostic strings.
Conf files with a structured format (yaml, json, ...) can help keep complexity at bay and are also more intuitive for users and admins: Well-structured configuration files reflect the architecture of the application.

Another problem with env vars is, that they often end up in the configuration files of other applications, e.g. docker-compose, ... For some config options that is convenient (e.g. database settings are well suited for the Compose file), for others not so much.

In practice I personally favour a combined approach, where (at least some) configuration file values can be overwritten using env vars. So I think the current mix is quite good: Database settings, debug mode, etc. in env vars, more persistent settings only in promgen.yml.

PR #43, opened 2018-01-29
Update prometheus.yml

Tommassino:
Change rule path to be consistent with https://github.com/line/promgen/blob/master/promgen/tests/examples/promgen.yml
Change regex to include all exporters

### DIFF
--- a/docker/prometheus.yml
+++ b/docker/prometheus.yml
@@ -21,7 +21,7 @@ scrape_configs:
         - "/etc/prometheus/promgen.json"
     relabel_configs:
     - source_labels: [__shard]
-      regex: *
+      regex: .*
       action: keep
 
   - job_name: 'blackbox'
--- a/docker/prometheus.yml
+++ b/docker/prometheus.yml
@@ -6,7 +6,7 @@ global:
     cluster_name: 'promgen'
 
 rule_files:
-- "/etc/prometheus/promgen.rule.yml"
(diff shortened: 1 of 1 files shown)


docker/prometheus.yml
```diff
@@ -21,7 +21,7 @@ scrape_configs:
         - "/etc/prometheus/promgen.json"
     relabel_configs:
     - source_labels: [__shard]
-      regex: docker-demo
+      regex: .*
```
kfdm: Instead of doing a regex .*, could actually drop the entire section to have Prometheus read in all the rules. I think it would be best to leave it there commented out as an example so maybe I'll give this a shot :)

docker/prometheus.yml
```diff
@@ -6,7 +6,7 @@ global:
     cluster_name: 'promgen'
 
 rule_files:
-- "/etc/prometheus/promgen.rule.yml"
+- "/etc/prometheus/promgen.rule"
```
kfdm: Ah, this is also slightly tricky because the extension depends on if you're running Prometheus 1.x or 2.x. I think docker-compose should probably default to running 2.x by default, but perhaps I'll put a comment here

kfdm:
I've created my own PR to hopefully add better inline comments to explain what it's trying to do there. Let me know if that helps answer the confusion around these settings.

Issue #41, opened 2018-01-29
promgen compose


Tommassino:
Hi,

promgen looks exactly what i need to run prometheus since I want to change the configurations of alerts and nodes dynamically. I've started with taking your docker-compose that you have in your project and altering it to include both the mysql and redis databases (later on I can move them outside of this docker). I can successfully start the docker containers and all of them seem to be running just fine, but I'm having issues configuring promgen.

I add a exporter successfully to the UI and the test functionality confirms it working, however the Prometheus server does not get it added to its targets. Further when I attempt to add a alert rule i get internal server errors (and nothing visible in the promgen log). What am I doing wrong?

PS: i ran the DB initialization scripts manually (i plan to run them automatically later).

![image](https://user-images.githubusercontent.com/698950/35500060-56ac0692-04d5-11e8-98f1-10bf3195e67d.png)

The prometheus server has been added:

![image](https://user-images.githubusercontent.com/698950/35500208-f2fea5ae-04d5-11e8-8fcb-d3b8f17a96a0.png)

The compose (WIP) used:

```
version: '2'
services:
  redis:
    image: redis
    container_name: redis
    network_mode: bridge
    restart: always
    ports:
      - 6379:6379

  mysql:
    image: mariadb
    container_name: mysql
    network_mode: bridge
    restart: always
    ports:
      - 3306:3306
    environment:
      MYSQL_ROOT_PASSWORD: secret
      MYSQL_DATABASE: pypromgen
      MYSQL_USER: promgen
      MYSQL_PASSWORD: promgen

  base:
    build: .
    image: promgen
    network_mode: bridge
    command: echo ""
    environment:
      CELERY_BROKER_URL: redis://redis:6379/0
      DATABASE_URL: mysql://promgen:promgen@mysql:3306/pypromgen
      SECRET_KEY: asdf

  prep:
    extends: base
    container_name: prep
    command:  sh -c "promgen migrate"
    links:
      - mysql
      - redis
    environment:
      DEBUG: 1

  web:
    extends: base
    container_name: web
    command: promgen runserver 0.0.0.0:8000
    links:
      - mysql
      - redis
      - prep
      - nodeexporter
    ports:
      - "8000:8000"
    environment:
      DEBUG: 1

  worker:
    extends: base
    container_name: worker
    command: worker -l info --queues localhost,celery
    links:
      - prometheus
      - redis
    volumes:
      - ./docker:/etc/prometheus

  prometheus:
    image: prom/prometheus
[...]

Issue #36, opened 2018-01-26
[Request] Promgen Tutorial


hmmxp:
Dear Developer,

Can provide a simple tutorial on how to use promgen? 
Also How Can We Configure Promgen When Prometheus is on a different server from Promgen?

kfdm:
You're right. While I think I have mostly documented the installation steps, there is not a lot of information on how to use Promgen itself

While it could use more detail, it's mentioned on some of the docker instructions regarding a worker node

```
# Run Promgen web worker. This is typically balanced behind an NGINX instance
docker run --rm -p 8000:8000 -v ~/.config/promgen:/etc/promgen/ line/promgen:master

# Run Promgen celery worker. Make sure to run it on the same machine as your Prometheus server to manage the config settings
docker run --rm -v ~/.config/promgen:/etc/promgen/ -v /etc/prometheus:/etc/prometheus line/promgen:master worker
```

I will see what I can do to explain things better. Sorry for the delays.

hmmxp:
Dear Paul,

Hopeful more tutorial will be out as it will be beneficial as i have around
10 Prometheus server spread across multiple location and Promgen looks like
a good solution to manage the environment

hmmxp:
# Run Promgen web worker. This is typically balanced behind an NGINX instance
docker run --rm -p 8000:8000 -v ~/.config/promgen:/etc/promgen/ line/promgen:master

# Run Promgen celery worker. Make sure to run it on the same machine as your Prometheus server to manage the config settings
docker run --rm -v ~/.config/promgen:/etc/promgen/ -v /etc/prometheus:/etc/prometheus line/promgen:master worker

Correct me if i am wrong

1. I need to have a central promgen server
2. For every prometheus server or server running the prometheus exporter, i need to have both promgen web worker and celery worker running?

kfdm:
You can have your Promgen web worker running anywhere you want and then you need a single worker for each Prometheus instance. I'm going to try to update the documentation around that this week.

hmmxp:
Dear Paul,

Thank you for the feedback, will await your documentation and see how to finish configuring my promgen stack as i am not sure how to manage my multiple Prometheus server with promgen

PR #46, opened 2018-01-30
Updating docs for Promgen's different worker modes

kfdm:
This should help answer some questions in #36 and #41 regarding management of Promgen workers and Prometheus nodes


kfdm:
Going to go ahead and merge this for now, since it answers some recent questions, and I'll keep working to fill in the missing bits as questions come in.

PR #50, opened 2018-01-31
Add a createsuperuser command

Tommassino:
A command that enables automatic bootstrapping via env variables.

### DIFF
--- a/promgen/management/commands/create_superuser_with_password.py
+++ b/promgen/management/commands/create_superuser_with_password.py
@@ -1,0 +1,39 @@
+from django.contrib.auth.management.commands import createsuperuser
+from django.core.management import CommandError
+
+
+class Command(createsuperuser.Command):
+    help = 'Crate a superuser, and allow password to be provided'
+
+    def add_arguments(self, parser):
+        super(Command, self).add_arguments(parser)
+        parser.add_argument(
+            '--password', dest='password', default=None,
+            help='Specifies the password for the superuser.',
+        )
+        parser.add_argument(
+            '--preserve', dest='preserve', default=False, action='store_true',
+            help='Exit normally if the user already exists.',
+        )
+
+    def handle(self, *args, **options):
+        password = options.get('password')
+        username = options.get('username')
+        database = options.get('database')
+
+        if password and not username:
+            raise CommandError("--username is required if specifying --password")
+
+        if username and options.get('preserve'):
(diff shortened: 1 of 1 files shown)


kfdm:
Thanks for the PR!
I do have a slight reservation about adding a command that duplicates a lot of the built in 'createsuperuser' command especially since it's typically something that would only happen a single time.

What is the use case for bootstrapping this way? For having a self contained docker image that you can add default passwords? (similar to setting them in the mysql docker image)

I think I would slightly prefer handling it Grafana style with a default admin account on first run, but I'd like to hear your thoughts :)
(partly because this would be less code to maintain and a bit simpler in some ways :) )

Tommassino:
My usecase for the most part is in this commit - https://github.com/Tommassino/promgen/commit/890e5a5eab5e5a5bf849ab9b9478b6fa22257ce9 in which i bootstrap the inital setup via environment variables. If there was a default admin account and an ability to change its password via the UI that would be possibly a better way of handling this, but as far as I am aware there is no user management implemented in promgen at the moment.

kfdm:
Promgen is (mostly) a standard django application so several things can be edited through the /admin/ page (after you create a superuser that can login of course). The django admin lets you change passwords.

As for users setting their own password, right now, we use Oauth to login so I haven't exposed that, but it should be easy to add a page for user's to set their password

Tommassino:
Ah, Im not really familiar with django and had no idea about the /admin page

kfdm:
How about this, since I'm a little apprehensive about adding a command just to manage the first user we can add a default user in bootstrap ``if user.count == 0 user.create(name='admin', . .. `` with the prompt warning the user to login and update?

hmmxp:
My suggestion is same as Paul, add a default user and upon completion of the setup and when user first launch the Promgen web application it will prompt the user to change the default password for the created default user admin.

I think this will lessen the steps and will provide a better experience

Tommassino:
I don't like the idea of a prompt on the first startup of promgen, this makes automatic deployment even more cumbersome. Ill investigate into the suggestion by kfdm where a default user gets created in the bootstrap process.

kfdm:
@Tommassino I created #59 to illustrate my idea for how to bootstrap a user without adding an additional command. Let me know what you think and if that would solve your use case.

kfdm:
I'm going to close this for now, since I believe I have a better solution ready in #59. May take another day or two to be ready though.

PR #54, opened 2018-02-03
Update instructions in CONTRIBUTING.md

seoester:
This pull request updates the "Setting up Promgen for development" section in the CONTRIBUTING file.

I added some commands initialising the required settings and also switched from `virtualenv` to the `venv` module.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,13 +15,14 @@ python3 -m venv /path/to/virtualenv
 source /path/to/virtualenv/bin/activate
 pip install -e .[dev]
 pip install mysqlclient # psycopg or another database driver
-# Provide initial configuration
+# Boostrap initial configuration
 promgen bootstrap
 # Enable DEBUG (and development) mode
 echo 1 > ~/.config/promgen/DEBUG
-# Copy default settings (amend by hand after)
-cp promgen/tests/examples/promgen.yml
-# Setup database and update tables
+
+# Amend bootstrapped configuration by hand: ~/.config/promgen/promgen.yml
+
+# Setup or update database schemas, create initial user and shard
 promgen migrate
 # Run tests
 promgen test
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -11,10 +11,16 @@ First of all, thank you so much for taking your time to contribute! We always we
...
+python3 -m venv /path/to/virtualenv
+source /path/to/virtualenv/bin/activate
 pip install -e .[dev]
 pip install mysqlclient # psycopg or another database driver
+# Provide initial configuration
+promgen bootstrap
+# Enable DEBUG (and development) mode
+echo 1 > ~/.config/promgen/DEBUG
+# Copy default settings (amend by hand after)
+cp promgen/tests/examples/promgen.yml
```
kfdm: This should be copied by bootstrap already, though perhaps I could have it automatically open up in $EDITOR if you thought that would be useful

CONTRIBUTING.md
```diff
@@ -25,6 +31,11 @@ promgen runserver
 
 > Note: Promgen strives to be a standard Django application. Make sure to apply standard Django development patterns.
 
+#### Custom Configuration Directory
+
+By default `promgen` uses `~/.config/promgen` as its configuration directory.
+This can be changed by setting the environment variable `CONFIG_DIR` to an alternative location whenever calling `promgen ...` commands.
```
kfdm: Perhaps I should change this to PROMGEN_CONFIG? Any thoughts on that?

CONTRIBUTING.md
seoester: Thanks for the feedback! You're right, bootstrap will already have written a config file.
Do you know if there is a standard command on UNIX systems to start the default editor? I know debian provides `editor` which maps to nano, vim, emacs, ... but I'm not sure about other distributions.

kfdm: EDITOR is the one I have seen that is most common though sometimes I have seen VISUAL. Trying to find a more definitive source but it can be difficult with things like this :)

seoester: Alright, perhaps a comment is feasible instead of dissecting UNIX standard practices ;)
On my system `$EDITOR` is unset when starting the shell and the `editor` executable references nano.

kfdm: Yeah, it's just a random thought I had. Feel free to keep things simple :)

kfdm:
Hmm, not sure why I missed reviewing this again. I'll go ahead and merge this, and then update README to point to this file to avoid the duplicate content. Thank you :)

PR #55, opened 2018-02-03
Enable debug mode when environment variable DEBUG is set

seoester:
The change allows setting the environment variable DEBUG to any value except '0' to enable the django debug mode.
Placing a non-empty file `DEBUG` in the `CONFIG_DIR` has the same effect (the file is loaded by the `envdir` module).

Note: Previously it was possible to enable the debug mode by leaving the `DEBUG` file empty. This would **not** be possible anymore after merging this PR.

Personally, I think it is worth breaking this functionality for higher compatibility with `envdir` and to keep the code more maintainable.
However, if you'd prefer to keep this feature, please get back to me.

### DIFF
--- a/promgen/settings.py
+++ b/promgen/settings.py
@@ -40,7 +40,7 @@ if not SECRET_KEY:
     SECRET_KEY = get_random_string(50, 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.path.exists(os.path.join(CONFIG_DIR, 'DEBUG'))
+DEBUG = 'DEBUG' in os.environ and os.environ['DEBUG'] != '0'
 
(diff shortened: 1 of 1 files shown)


promgen/settings.py
```diff
@@ -40,7 +40,7 @@
     SECRET_KEY = get_random_string(50, 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.path.exists(os.path.join(CONFIG_DIR, 'DEBUG'))
+DEBUG = 'DEBUG' in os.environ and os.environ['DEBUG'] != '0'
```
kfdm: Could write this as 
```python
DEBUG = os.environ.get('DEBUG', '0') != '0'
```
To make it a bit simpler perhaps ?

kfdm:
Thank you :)

PR #56, opened 2018-02-03
Dockerfile Rework

seoester:
I re-wrote the `Dockerfile` to be more in line with common `Dockerfile`s.

The only significant difference is the user running promgen: I removed the promgen user, that means that the web app and the worker are now run as root. This simplifies the permissions required for the configuration files. Running applications in docker as root is a common strategy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,46 +1,36 @@
 FROM python:3.5.3-alpine
-MAINTAINER paul.traylor@linecorp.com
+LABEL maintainer=paul.traylor@linecorp.com
 
 ENV PROMETHEUS_VERSION 2.1.0
 ENV PROMETHEUS_DOWNLOAD_URL https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz
 
-RUN adduser -D -u 1000 promgen promgen
-RUN apk add --no-cache --update mariadb-dev build-base bash && \
-    rm -rf /var/cache/apk/*
-
-RUN set -ex; \
-        apk add --no-cache curl tar; \
-        curl -L -s $PROMETHEUS_DOWNLOAD_URL \
-		| tar -xz -C /usr/local/bin --strip-components=1 prometheus-${PROMETHEUS_VERSION}.linux-amd64/promtool; \
-        apk del curl tar; \
-	rm -rf /var/cache/apk;
-
 ENV PYTHONUNBUFFERED 1
 ENV PIP_NO_CACHE_DIR off
 
-RUN mkdir -p /etc/prometheus
-RUN mkdir -p /etc/promgen
-RUN mkdir -p /usr/src/app
-RUN chown promgen /etc/prometheus
-
 COPY docker/requirements.txt /tmp/requirements.txt
-RUN pip install -r /tmp/requirements.txt
-
 COPY setup.py /usr/src/app/setup.py
 COPY promgen /usr/src/app/promgen
 COPY promgen/tests/examples/promgen.yml /etc/promgen/promgen.yml
+COPY docker/docker-entrypoint.sh /
 
 WORKDIR /usr/src/app
-RUN pip install -e .
 
 ENV CONFIG_DIR=/etc/promgen
 
-USER promgen
+RUN set -ex; \
+	apk add --no-cache --update mariadb-dev build-base bash && \
+	apk add --no-cache curl tar; \
+	curl -L -s $PROMETHEUS_DOWNLOAD_URL \
+		| tar -xz -C /usr/local/bin --strip-components=1 prometheus-${PROMETHEUS_VERSION}.linux-amd64/promtool; \
+	apk del curl tar; \
+	rm -rf /var/cache/apk; \
+	mkdir -p /etc/prometheus; \
+	pip install -r /tmp/requirements.txt; \
+	pip install -e .; \
+	SECRET_KEY=1 promgen collectstatic --noinput;
+
 EXPOSE 8000
 
-RUN SECRET_KEY=1 promgen collectstatic --noinput
-
-COPY docker/docker-entrypoint.sh /
 VOLUME ["/etc/promgen", "/etc/prometheus"]
 ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD ["web", "--bind", "0.0.0.0:8000", "--workers", "4"]


Dockerfile
```diff
@@ -1,46 +1,36 @@
...
 COPY promgen /usr/src/app/promgen
 COPY promgen/tests/examples/promgen.yml /etc/promgen/promgen.yml
+COPY docker/docker-entrypoint.sh /
 
 WORKDIR /usr/src/app
-RUN pip install -e .
 
 ENV CONFIG_DIR=/etc/promgen
 
-USER promgen
```
kfdm: I have to admit I'm quite ignorant of docker best practices, but I thought it was generally discourated to run as root. I know Prometheus 2.0 was changed to not run as root ( https://github.com/prometheus/prometheus/issues/1637 ) though I think that also caused issues for some other things. Hmm.

seoester: You are right, it is not as simple as I wrote in the description.

It is theoretically not possible for the docker root user to break out (and gain host root privileges). In production, containers are usually run as a non-root user to add an additional layer of security. And to prevent attackers from compromising the (ephemeral) container fs, of course.

Using `nobody` as the user to execute commands would be good for security and does not pose an issue for the Promgen Web App. The worker, however, needs write access to an external / shared volume.
Adding a new user to a docker image would require very specific configuration on the host system: Creating a new isolated user with the same uid and granting access only to the docker volume mount point.
By using root the "self-contained" aspect of the docker image is preserved. In more advanced or production deployments the user will (and should) most likely be changed to a custom user by the operator.

seoester: I started reading up on the issue, as I'm not the most up-to-date either.

While most agree that it's best not to run as root, there is no consensus on what to do instead.
To me, it seems best to run under a random but static uid by default.

For Promgen the situation is slightly more complicated, as setting up write access to the external mount point is not optional but required for running the worker.

 * <https://github.com/prometheus/prometheus/issues/3441>
 * <https://github.com/minio/minio/pull/4877>
 * <https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user>

kfdm:
I also need to test a bit more, but can probably move the 'apk del' to the end to make the container a bit smaller

```patch
diff --git a/Dockerfile b/Dockerfile
index 733a9a2..75e5547 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,11 +22,12 @@ RUN set -ex; \
 	apk add --no-cache curl tar; \
 	curl -L -s $PROMETHEUS_DOWNLOAD_URL \
 		| tar -xz -C /usr/local/bin --strip-components=1 prometheus-${PROMETHEUS_VERSION}.linux-amd64/promtool; \
-	apk del curl tar; \
-	rm -rf /var/cache/apk; \
 	mkdir -p /etc/prometheus; \
 	pip install -r /tmp/requirements.txt; \
 	pip install -e .; \
+	apk del curl tar build-base; \
+	rm -rf /var/cache/apk; \
 	SECRET_KEY=1 promgen collectstatic --noinput;
 
 EXPOSE 8000
```

```
REPOSITORY                                TAG                 IMAGE ID            CREATED              SIZE
promgen                                   test                3425ff233702        About a minute ago   442MB
<none>                                    <none>              54c546587f72        21 minutes ago       588MB
```

seoester:
Nice savings :)
I'll add WIP to the title for now as to indicate that we're still actively amending this.

Dockerfile
```diff
@@ -1,46 +1,36 @@
...
-
 COPY setup.py /usr/src/app/setup.py
 COPY promgen /usr/src/app/promgen
 COPY promgen/tests/examples/promgen.yml /etc/promgen/promgen.yml
+COPY docker/docker-entrypoint.sh /
 
 WORKDIR /usr/src/app
-RUN pip install -e .
 
 ENV CONFIG_DIR=/etc/promgen
```
seoester: Amend to match #61 (`PROMGEN_CONFIG_DIR`)

seoester:
I rebuilt the container from the current master branch, and the size has been greatly reduced :+1: 

```
promgen-test                                                        latest              0c63338ab2a0        45 hours ago        195MB
```

seoester:
The primary open question is: What user should promgen run as? ([comments](https://github.com/line/promgen/pull/56#pullrequestreview-93844085))

The most secure option is to create a new user in the container with a random but static uid and gid. However, this poses a problem because the worker has to write files to an (external) docker volume. By running as such a user, operators have to configure appropriate permissions for the prometheus configuration files.

Some options:

 * Run as user root. Operators should change the user in production environments.
 * Run as user promgen (as above).
   * Describe how to set permissions in the "getting started with docker-compose" documentation.
   * In docker-compose.yml, an additional service could be used to properly set up file permissions. This service could also bootstrap the configuration completely, requiring no manual setup before `docker-compose up`.
   * Change the user of the worker service in docker-compose.yml to root.

kfdm:
Yeah, I do not have a good answer to that since I am quite unaware of the best practices. I think I will yield that decision to what you think is best :)

seoester:
I had some further ideas for a more holistic approach to the docker image, I introduced the following changes:

 * Removed bash, docker-entrypoint.sh now uses /bin/sh
 * Perform migrations whenever starting a web or web-dev container.
   * `migrate` is idempotent on consecutive executions :+1:
   * By setting `PROMGEN_DISABLE_MIGRATE` to `'1'` migrations are disabled (perhaps useful when running with multiple containers)
 * Removed the `docker-compose-bootstrap` command
 * By default: Use user promgen with id 16395.
 * Added a `init-config` command which runs `bootstrap` and automatically fixes permissions when executed as root user.
   * Added --no-input flag to `bootstrap` so it can be executed without asking the user for additional information
 * `collectstatic` and provide a mountable volume to serve from a separate container.
   * In Dockerfile, `collecstatic` to `STATIC_ROOT=/usr/src/app/static`
   * The files are copied over to the volume at `/srv/promgen/www` (`/static`) in the docker-entrypoint.sh file.
 * Enable overwriting STATIC_URL via env var

The aim is to have a simpler setup experience as well as to accomodate production setups.
Auto migrations are used to allow a docker image upgrade without requiring any manual intervention.

To do:

 * [ ] Prometheus: Write configuration file (docker-compose context)
 * [ ] Alertmanager: Write configuration file (docker-compose context)

Dockerfile
```diff
@@ -1,46 +1,39 @@
...
-ENV CONFIG_DIR=/etc/promgen
+ENV PYTHONUNBUFFERED=1 \
+	PROMGEN_CONFIG_DIR=/etc/promgen \
+	STATIC_ROOT=/srv/promgen/www/static
 
-USER promgen
-EXPOSE 8000
+RUN set -ex; \
+	addgroup -g 16395 promgen; \
+	adduser -D -u 16395 -G promgen promgen; \
```
kfdm: Is there something special about the value 16395 here ? Shouldn't be a problem, I'm just curious why you chose this value :)

docker/docker-entrypoint.sh
```diff
@@ -1,20 +1,100 @@
-#!/bin/bash
+#!/bin/sh
+
+function check_migrate() {
```
kfdm: Moving this to a function to reuse is a good idea 👍

docker/docker-entrypoint.sh
```diff
@@ -1,20 +1,100 @@
...
+
+    promgen migrate
+    promgen register "${PROMGEN_REGISTER_SHARD}" "${PROMGEN_REGISTER_HOST}" "${PROMGEN_REGISTER_PORT}"
+
+    set +x
+  fi
+}
+
+function copy_static() {
+  if [ ! -f /srv/promgen/www/static/.present ]; then
```
kfdm: Just reading the code, I'm not entirely sure the purpose for having both /srv/promgen/www and /usr/src/app/static. Is this so that one can mount the directory and then serve the files from nginx/apache or for some other reason?
I think a comment either in the file or the docker instructions will be necessary for clarification :)

docker/docker-entrypoint.sh
```diff
@@ -1,20 +1,100 @@
...
   # Shortcut to start a celery worker for Promgen
   set -- celery "-A" promgen "$@"
   ;;
-web)
-  # Shortcut for launching a Promgen web worker under gunicorn
+init-config)
   shift
-  set -- gunicorn "promgen.wsgi:application" "$@"
+
+  exec promgen bootstrap --noinput "$@"
```
kfdm: I could be mistaken, but I *think* using `exec` here means that this shell will be replaced by the running program so the later steps will not run.

seoester: I meant to check this, you're absolutely right, exec just wraps the corresponding syscall.

docker/docker-entrypoint.sh
seoester: Yes, the reason is the docker volume management: Every time the container is created, a new volume is allocated and mounted at the path `/srv/promgen/www`. This means all content in the directory (e.g. created in the Dockerfile) will be invisible.
So instead, on the start of the web app, the files are copied across, if they don't exist yet. The `.present` file represents a minor security consideration: When not forbidden by the web server, the file will be accessible by anyone with HTTP access. I decided to only run `collectstatic` in the Dockerfile, just in case it involves some computationally complex work.

Other containers can then mount the volume (e.g. with the `--volumes-from` flag, by mounting a custom volume in both containers or by using the bind volumes).

Dockerfile
seoester: There is nothing special about this value, it is an arbitrary uid >> 1000. This is the random but static uid I wrote about before.

Some projects choose to simply run `adduser` as the first command and hope for the same uid to be allocated every time. For promgen, the uid is of the upmost importance, because if permissions are misconfigured, the worker will be unable to write any configuration files. That's why I think explicitly setting the uid is a good idea.

Because the uid is much greater than 1000, it is unlikely to map to any user on the local system.

kfdm: Ok. Should be fine, I just wish there was more authoritative sources about values like this 😓 
I agree blindly trusting `adduser` is scary.
For my own deployment I use 9090:9090 for Prometheus (to match the default port) but I suppose in Promgen's case we can't easily do something similar.

docker/docker-entrypoint.sh
kfdm: One consideration instead of adding a new file (.present) perhaps can just test for a file that we expect to be there, `/srv/promgen/www/static/js/promgen.js` otherwise, seems like a reasonable explanation :)

seoester: That sounds good :+1:

seoester:
I made the changes we discussed and also removed `/etc/promgen` as a volume. Operators can still choose to mount a volume there, but I think it doesn't make sense to have it declared in the Dockerfile.

Still have to properly test everything though.

seoester:
Some of the ideas implemented in the PR are still valid but must be re-applied against the current version. I'll re-apply next week.

PR #57, opened 2018-02-03
[WIP] Docker Compose Update

seoester:
This PR contains new two new `docker-compose.yml` files, one for the backend (mariadb and redis) and one for Promgen, Prometheus, etc.

The docker docs are also updated to describe how to use Docker Compose to quickly get started with promgen.

I split the `docker-compose.yml` file into two so that the backend can be started separately, and only when ready promgen is started. A better solution would be to integrate wait-for-db functionality into the promgen image.
However, I have not looked into this yet.


kfdm:
I suppose my biggest question (again due to my lack of knowledge regarding docker best practices) but I'm a bit unsure why to split things up into two docker files. My initial impression was that the docker-compose file gave a base starting point as a demo or to get started but for most production deployments it would always require some adjustments. It's unfortunate that docker does not have a built in way to wait-for a container and Django also does not have a better way to have it reconnect so perhaps keeping it as a single docker-compose file and including a wait-for-it type check may be the way to go.

Previously I used the sentry onpremise script as a base, since Sentry is also a django application
https://github.com/getsentry/onpremise/blob/master/docker-compose.yml

Looks like there may be a default way built into newer docker versions but I have not tested it myself
https://stackoverflow.com/questions/31746182/docker-compose-wait-for-container-x-before-starting-y/41854997#41854997

seoester:
I agree that applications (and docker containers as examples for microservices especially) should be resilient against failure of external dependencies / services.

I've read both about django supporting database restarts and django stalling when the database is shut down and then started again.
I would like to test this out, but might not get around to it this week. Until then, I'd like to keep this PR open.

By splitting into two files the backend can be treated as an external service which is assumed to always be available.

`HEALTHCHECK` is not an adequate method for waiting on other containers, it is meant for monitoring and automated life-cycle management.

seoester:
This would be much improved by pending PR #95.

docker/docker-entrypoint.sh
```diff
@@ -16,6 +16,17 @@ web)
   shift
   set -- gunicorn "promgen.wsgi:application" "$@"
   ;;
+wait-runserver)
```
kfdm: Instead of using a separate wait-runserver block, I wonder if this would be better added to the existing 'web' block above here. Runserver is sufficient for development but perhaps the docker container should use gunicorn.

kfdm: I realized that my Dockerfile entrypoint uses *web* though my original docker-compose uses *runserver*. It's probably best to make both as *web* since *runserver* is mostly for development use. What do you think?

I also thought a bit about having the `promgen check ; sleep` check apply to all of the commands by putting it at the top, but there are also likely valid cases or not forcing that check while debugging the system. Do you have any thoughts about this as well?

seoester: gunicorn is running a WSGI server, so this would require an additional container running nginx (or similar) to be able to serve HTTP requests.

I think it is sensible to have two different entrypoints, both with a wait mechanism:

 * An entrypoint starting up a HTTP server using the runserver command .
 * An entrypoint starting up a WSGI server using gunicorn.

I think commands should be regarded individually when considering whether to wait for db. With one-off commands run manually, it is in the user's interest to get an immediate response. For example `worker` does not really need to wait for db, as it's basic functionality does not depend on the availability of mysql and resilience to redis outages is already built into celery.

Perhaps for some one-off commands it would be useful to have optional wait for db functionality, for example when running it from a docker-compose file. In this case either a new command `wait-…` could be introduced using case fallthrough or an env var could be checked to decide whether to wait for db or try immediately.

docker-compose.yml
```diff
@@ -1,74 +1,86 @@
...
+  promgen-setup:
+    image: line/promgen
+    depends_on:
+      - mysql
+    environment:
+      PROMGEN_REGISTER_SHARD: docker-default
+      PROMGEN_REGISTER_HOST: prometheus
+      PROMGEN_REGISTER_PORT: 9090
+      DATABASE_URL: mysql://root:secretrootpassword@mysql:3306/promgen
+      SECRET_KEY: some-secret-key-AhvSh+xGkm9kA,ghfi4?w8K
```
seoester: I'm not familiar with django commands and best practices, should this stanza also include `DEBUG: 1`, the `CELERY_BROKER_URL`? Is `SECRET_KEY` even required?

docker/docker-entrypoint.sh
kfdm: gunicorn can speak HTTP so nginx is not required though it is often used in a production environment. I do not have a strong opinion since I tend to consider docker-compose a bit more of a demo than a fully ready production environment. When deploying for real it's likely that most would want to tweak their setup a bit.

Maybe best to keep things simple in this pass. Can always add additional wait- commands or make it global if required in the future.

docker-compose.yml
kfdm: DEBUG would only be necessary for times when you want extra debugging information, typically just for development. Could drop it from the docker-compose file and be fine.

SECRET_KEY in theory is only required for http but I believe the django startup code errors out if it is not set somewhere. Could set a default SECRET_KEY again in settings.py so that we could remove it from the docker-compose file.

docs/docker.rst
```diff
@@ -1,17 +1,75 @@
 Running with Docker
 ===================
 
+The easiest way to get started is by using
+`docker-compose <https://docs.docker.com/compose/>`__.
+
+Promgen ships with two ``docker-compose.yml`` files:
```
kfdm: Since the two files were combined, I think these docs need to also be updated :)

seoester: Yeah, true :smile:  I'll update the documentation when this PR is about to settle.

docker/docker-entrypoint.sh
seoester: Hm, what do you think about these entrypoints, both protected by `promgen check`:

 * `web` - gunicorn running the application. Used as the default for docker-compose.
 * `web-dev` - django `runserver`. Meant to be used for testing and development.

`web` could then be used for a more production-y setup together with nginx. Additional documentation could be compiled on how to achieve this (e.g. using two containers sharing a volume for hosting of the static files).

kfdm: Sounds reasonable. I think those names are more descriptive than *wait-runserver* for those who may not be as familiar with django directly.

seoester:
I'm interested in revisiting this, as we are finally deploying Promgen in a Docker Swarm.
I'm closing this now as the numerous changes to Promgen and new Docker versions require a re-write of the compose file.

Issue #49, opened 2018-01-31
 怎么定义prometheus规则 How to define the Prometheus rule


peixiaoqing:
HI  developer：
When I use the test rules, the backstage is wrong, 500, how do you solve it in general? How do you use promgen to generate Prometheus rules?
![qq 20180131134955](https://user-images.githubusercontent.com/35130532/35607275-4a1e4d8a-068e-11e8-923b-b91719195e49.png)

Tommassino:
May be related to https://github.com/line/promgen/pull/39 
Try running the command `promgen targets`

peixiaoqing:
Thanks to the developers to reply to the news in their busy schedule
According to the command to my reply "promgen targets" tried to create the rules, the background or the 500 false, I can see the bottom of the screenshot.
Now I say the way I install it, I install it according to the installation of centos7, and do not use docker. After the installation is complete, I add the host and monitor port, test can pass, but when adding rules, place error 500 rules: node_load1{project= 34, <exclude>} > 5 is my rule is wrong, which is wrong, please help out, thank you。
![aa](https://user-images.githubusercontent.com/35130532/35657785-44e2fad6-0739-11e8-9da3-a8abf90e36d6.png)
![bb](https://user-images.githubusercontent.com/35130532/35657792-4a6478c2-0739-11e8-9baf-a45843181f7e.png)
![cc](https://user-images.githubusercontent.com/35130532/35657827-7e7984c2-0739-11e8-9d2e-85287ebd3869.png)
![dd](https://user-images.githubusercontent.com/35130532/35657832-8105932a-0739-11e8-9309-ce40c097afe0.png)

kfdm:
The rule tester is still probably a bit rough. If you go into the Promgen Admin and look at the Shard list ( /admin/promgen/shard/ ) find the shard you're using and ensure "Proxy" is checked.

That whole section is poorly documented so I need to fix that. Sorry for the confusion :(

peixiaoqing:
Hello developer: according to the tips given to me above, I don't see very well. Can you give me a document on the definition of the rules? Thank you

peixiaoqing:
Hello developer: according to the tips given to me above, I understand very well. Can you give me a document on the definition of the rules? Thank you

peixiaoqing:
Hello developer: according to my tips above, I don't understand very well. Can you give me a document on the definition of the rules? Thank you

kfdm:
[...]

PR #59, opened 2018-02-06
Add default admin user if no superusers exist on bootstrap

kfdm:
Prompted by @Tommassino in #50 this adds a default admin user on initial bootstrap


promgen/management/commands/bootstrap.py
```diff
@@ -58,3 +59,14 @@ def handle(self, **kwargs):
         # Schemes based on list of supported brokers
         # http://docs.celeryproject.org/en/latest/getting-started/brokers/index.html
         self.write_setting('CELERY_BROKER_URL', test=URLValidator(schemes=['redis', 'amqp', 'sqs']))
+
+        if User.objects.filter(is_superuser=True).count() == 0:
```
kfdm: I just realized that bootstrap is often run before migrate so this would cause problems. I may need to have bootstrap automatically call migrate which would help test the settings and ensure that we're able to create the super user.

seoester:
Thank you for considering simplifying application bootstrapping :)

I think the dilemma is, that the bootstrap command so far only handled the static configuration. This is another way of looking at this:

 1. Configuration Boostrapping (`bootstrap`)
 2. Database Initialisation (`migrate`)
 3. Application Bootstrapping: Default user, ... (…)

For shipping applications as docker images / packages the last step is of special interest.

kfdm:
Breaking it into those three steps does illustrate the problem much better. I suppose at the same time we bootstrap the initial login user, also should bootstrap the initial shard which is related to the rule test not working as expected in #49

kfdm:
Ok, I think I have an idea for a clean way to handle this.

I will revert some of the changes in bootstrap and leave that as ONLY modifying the settings (env/file)
I can then take advantage of the post_migrate hook to run a check after 'promgen migrate' is run, to see if an admin user and default shard need to be created. I'm testing some of this locally, but I also want to fix and squash the migrations to hopefully speed up part of tests and new installations.

https://docs.djangoproject.com/en/2.0/topics/migrations/#squashing-migrations
https://docs.djangoproject.com/en/2.0/ref/signals/#post-migrate

kfdm:
@Tommassino and @seoester this should break up bootstrap into two steps

1. promgen bootstrap - Will configure the file and env settings required
2. promgen migrate - Will bootstrap the database and any default data that needs to be generated.

Let me know if this seems reasonable

Tommassino:
From what i see in the bootstrap command, it is interactive and adds the default configurations etc. I think this works just fine for manual setups.

In the current promgen all that was missing in order to completely automate the setup was that a superuser could not be created non-interactive and that is solved by this PR. An additional bootstrapping command seems useful for manual setups, which are probably more important than my automated deployment use-case anyway. Nevertheless I appreciate you adding a default superuser so that i can deploy promgen automatically :)

Issue #58, opened 2018-02-04
Promgen Worker


hmmxp:
Dear All,

Currently i have created an installation guide for Promgen Web on CentOS 7, how can i install and configure Promgen Worker On CentOS 7?

seoester:
The preparation and installation required for the worker are the same as for the web app.

Then you register and start like this:

```sh
# Register prometheus server
# Syntax: promgen register <shard name> <prometheus host> <prometheus port>
promgen register default localhost 9090
# Start worker
# Syntax: celery -A promgen worker -l <log level> -Q <queues>
celery -A promgen worker -l info -Q localhost
```

The worker needs to subscribe to the queue with the host name of the prometheus server. 

You can also have a look at #57, it includes documentation on how to setup the web app and the worker using docker.

hmmxp:
i believe the above is for push model if promgen and promethues is installed on the same server.

But my promethues and promgen is on different server, how can i configure pull model?

kfdm:
The celery worker is installed the same way as the web app and as long as they can reach the same database and redis instance, it doesn't matter if the web app and workers are on different servers.

Apparently I forgot to merge some docs updates but you should be able to see some illustrations here
https://github.com/line/promgen/pull/46/files
I'll try to go through the PRs on Monday.

hmmxp:
Dear Paul & Seoster,

Will await the documents and finish testing in my linux environment and if i have spare time will test on docker

thank you all for the wonderful job

kfdm:
Correct. Those commands register the worker in Promgen's database so the data is saved between reboots and anything else :)

The worker flags you need to add to your startup command so that it knows which queue to run from.

kfdm:
uWSGI would be used to run the Promgen worker, so there are no worker flags required.

Something like systemd or another process manager would run the worker on the same node as Prometheus. I use something like this when deployed from Ansible


```
[Unit]
Description=Daemon for Celery
After=network.target

[Service]
Type=simple
ExecStart={{ promgen_virutalenv }}/bin/celery -A promgen worker -l info --queues {{ ansible_host }}
Restart=on-failure
User=prometheus

[Install]
WantedBy=multi-user.target
```

hmmxp:
Dear Paul,

Noted on this let me test it out and feedback and update the document as needed

PR #35, opened 2018-01-19
Create Installation Guide

hmmxp:

PR #60, opened 2018-02-06
More documentation for shards and web nodes

kfdm:
While reviewing the guide @hmmxp wrote in #35 and reflecting on the question in #58 I realized that the types of workers are poorly defined and need more clarification.

There are three kinds of processes.

Django application that provides the UI/API
Celery worker that updates Prometheus settings
Celery worker that sends out Alerts

Roughly the three processes can be run in this way (in development for illustration purposes)

```bash
# Promgen Web Node (Django)
promgen runserver
# Promgen Alert Worker (Celery)
# Not specifying --queues will have Celery listen to the default queue which is used for alerts
celery -A promgen 
# Promgen Config Worker (Celery)
# Workers running on the Prometheus node should specify the same value as stored in Promgen's database
celery -A promgen --queues <hostname> 
```


hmmxp:
Dear Paul,

Thank you for clarifying the above, please correct me if i am wrong

Promgen Web Server
- Just need to ensure the web server is running

Prometheus Server
- Need to have promgen web application running and celery running via systemd or just need the celery running?

kfdm:
Reordering the documentation to better explain the three different types of processes required to run Promgen and adds a section for documenting how to use Promgen

PR #61, opened 2018-02-06
Updates to CONFIG_DIR, and STATIC_ROOT

kfdm:
* Rename CONFIG_DIR to PROMGEN_CONFIG_DIR to be more descriptive
* Add PROMGEN_CONFIG as the path to promgen.yml
* Add STATIC_ROOT (was unintentionally hardcoded)
* Move SECRET_KEY within settings.py to allow override in promgen.yml

STATIC_ROOT was missed and I've updated PROMGEN_CONFIG_DIR based on a comment in #54

@seoester let me know if this seems reasonable to you regarding your documentation changes.

### DIFF
--- a/promgen/__init__.py
+++ b/promgen/__init__.py
@@ -5,8 +5,8 @@ import os
 import envdir
 
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'promgen.settings')
-os.environ.setdefault('CONFIG_DIR', os.path.expanduser('~/.config/promgen'))
+os.environ.setdefault('PROMGEN_CONFIG_DIR', os.path.expanduser('~/.config/promgen'))
 default_app_config = 'promgen.apps.PromgenConfig'
 
-if os.path.exists(os.environ['CONFIG_DIR']):
-    envdir.open(os.environ['CONFIG_DIR'])
+if os.path.exists(os.environ['PROMGEN_CONFIG_DIR']):
+    envdir.open(os.environ['PROMGEN_CONFIG_DIR'])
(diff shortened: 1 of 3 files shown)


hmmxp:
Dear Paul,

Much appreciated as this update will resolve the static file issues not displaying correctly and i no need to manually move the static file and configure alias in nginx

seoester:
Looks good to me, I'll amend my PR to match the new `PROMGEN_CONFIG_DIR` setting.

My only concern is, that there are now three ways to specify `SECRET_KEY`: The env var, the `SECRET_KEY` file and `django.SECRET_KEY` key in the promgen config file.

kfdm:
The env var and the SECRET_KEY file are basically the same. Everything in PROMGEN_CONFIG_DIR that is in ALLCAPS is loaded into the environ dictionary. I agree having multiple ways to configure things is a little strange, so perhaps this should be re-visited to clean it up a bit more.

At a high level, I suppose I imagined an org might build their own Docker image with their own promgen.yml file baked in, and then just use ENV vars to set connection information DATABASE and CELERY_BROKER_URL via ENV vars

seoester:
Alright, I see.
What are your thoughts on precedence? As I understand the code `django` keys override env vars, which is counter-intuitive at least for a docker deployment.

kfdm:
You're right >_<
I think env should override the config variables so I'll need to take another pass at it.

kfdm:
I want to go ahead and merge this to fix the variable naming and missing setting for STATIC_ROOT knowing that I still have the precedence incorrect. I will file a new PR to fix the precedence

PR #71, opened 2018-03-15
await mysql when promgen starts

m-tkg:
web container for promgen migrate must await not only mysql-container but also mysql-server starts.
so I implemented waiting loop using mysqladmin.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
   base:
     build: .
     image: promgen
-    command: sh -c "promgen migrate && promgen register docker-demo localhost 9090"
+    command: sh -c "until mysqladmin ping -h mysql --silent; do sleep 3; done; promgen migrate && promgen register docker-demo localhost 9090"
     environment:
       CELERY_BROKER_URL: redis://redis:6379/0
       DATABASE_URL: mysql://promgen:promgen@mysql:3306/promgen
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apk add --no-cache --update mariadb-dev build-base bash && \
     rm -rf /var/cache/apk/*
 
 RUN set -ex; \
-        apk add --no-cache curl tar; \
+        apk add --no-cache curl tar mysql-client; \
         curl -L -s $PROMETHEUS_DOWNLOAD_URL \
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
   base:
     build: .
     image: promgen
-    command: sh -c "promgen migrate && promgen register docker-demo localhost 9090"
+    command: sh -c "until mysqladmin ping -h mysql --silent; do sleep 3; done; promgen migrate && promgen register docker-demo localhost 9090"
```
kfdm: This came up when talking with @seoester in #57 though I'm admittedly unfamiliar with docker best practices. I've seen other articles that talk about using 'wait-for' to see when the database is ready

https://github.com/eficode/wait-for
https://docs.docker.com/compose/startup-order/

Trying to fit everything on a single line starts getting a bit messy, so perhaps it would be better to add this to docker-entrypoint.sh ?

```sh
docker-compose-bootstrap)
  # Shortcut to wait for Mysql to startup and run migrations
  until mysqladmin ping -h mysql --silent
  do
    sleep 3
  done
  promgen migrate
  promgen register docker-demo
  ;;
```

And then we could just make this line
```
command: docker-compose-bootstrap

seoester: Unfortunately, I still haven't gotten around to testing out django's behaviour in case of an unavailable database :/

As to this issue: I agree that waiting behaviour should be part of the docker image and thus should be included in the docker-entrypoint.sh.
Instead of / Additionally to creating a new command, the waiting mechanism should also be included in the `migrate` and `register` commands (really any command that requires a database connection). This way, the image behaves consistently.

Two more thoughts:

 * It might be sensible to have an upper bound on the number of attempts to reach the database.
 * Perhaps there is a way to use python + django to 'try' the database? That way, one additional dependency is avoided and the 'try' uses the same logic as the commands themselves.

kfdm: Tested this briefly today, it may make more sense to use one of the built in Django commands so you could run something like `promgen check` or `promgen inspectdb`. One advantage of this would be that it would be database agnostic 

1: https://docs.djangoproject.com/en/1.11/ref/django-admin/#check
2: https://docs.djangoproject.com/en/1.11/ref/django-admin/#django-admin-inspectdb

kfdm:
Fixed in #92

PR #92, opened 2018-07-18
Fix some version dependencies for docker container

kfdm:


docker/docker-entrypoint.sh
```diff
@@ -7,6 +7,17 @@ if [ "${1:0:1}" = '-' ]; then
 fi
 
 case "$1" in
+docker-compose-bootstrap)
```
kfdm: This should handle the same cases as #71

seoester:
Great stuff! This should be really useful for compose and docker in general.

I had the idea of making the parameters configurable via env vars. I'll create a PR the upcoming days.

PR #114, opened 2018-11-28
Support adding exporter defaults from admin gui

kfdm:
* Migrate from using a settings file to register default exporters to managing it via the django admin to ease management tasks.
* Provide a `register-exporter` CLI subcommand for use from configuration management
* Rename `register` to `register-server` to distinguish commands


m-tkg:
I found `Makefile` contains some tabs after `.PHONY`.

PR #124, opened 2019-02-06
[CHANGES] Rewritten display of alerts and silences to use vuejs

kfdm:
* Initial work to move from jquery to vuejs for more straightforward code
* Fetching alerts and silences uses javascript fetch api
* More client side rendering results in smaller requests
* Uses datetime-local for datepicker. Currently this only works at chrome
  but will consider looking at something for firefox and chrome support
  in the near future

There is a slight regression in the since that Firefox and Safari do not properly support `datetime-local` for `input` fields but since this PR has already gotten rather large, they will be handled in a future pull request


promgen/templates/promgen/alert_row.html
```diff
@@ -0,0 +1,21 @@
...
+<td>
+    <a :href="alert.generatorURL" v-text="alert.startsAt"></a>
+</td>
+<td>
+    <dl class="dl-horizontal">
+        <dt>Labels</dt>
+        <dd>
+            <ul class="list-inline">
+                <li v-for="(value, label, index) in alert.labels">
+                    <a @click.prevent="silenceAppendLabel" class="label label-warning" v-bind:data-label="label" v-bind:data-value="value">{{label}}:{{value}}</a>
```
ktarow: Since v-on is omitted, it is better to omit v-bind.

promgen/templates/promgen/silence_row.html
```diff
@@ -0,0 +1,16 @@
+{% verbatim %}
+<td>{{silence.endsAt}}</td>
+<td>
+    <ul class="list-inline">
+        <li v-for="match in silence.matchers">
+            <a @click.prevent="silenceAppendLabel" class="label label-warning" v-bind:data-label="match.name" v-bind:data-value="match.value">{{match.name}}:{{match.value}}</a>
```
ktarow: same as `promgen/templates/promgen/alert_row.html#L11`

promgen/templates/promgen/alert_row.html
kfdm: `v-on` has a shortcut as `@` so `@click.prevent` is the same as `v-on:click.prevent`. The `v-bind:data-label` are so I can read those values from the click event

ktarow: I know that v-bind also has a shortcut.

https://vuejs.org/v2/guide/syntax.html#v-bind-Shorthand

kfdm: Ah, yes I should probably be more consistent with using short forms or not. Do you have a preference for short vs long forms ?

ktarow: I like short form better.
However, if the long form is used in the project, we will use that.
Short form was used in this PR, so I wrote this comment.

promgen/templates/promgen/global_messages.html
```diff
@@ -0,0 +1,8 @@
+{% for message in messages %}
+<div class="alert {% if message.tags %}alert-{{ message.tags }}{% endif %}" role="alert">
```
Asuforce: ref: https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax
This is a little difficult to understand, so how about this ?

```suggestion
<div class="alert" :class="{ alertMessageTags }" role="alert">
```

and you can write such a property.

```js
computed: {
  alertMessageTags: function () {
    return {
      `alert-${message.tags}`: this.message.tags
    }
  }
}
```

promgen/static/js/promgen.vue.js
```diff
@@ -0,0 +1,130 @@
...
+        filterActiveSilences: function () {
+            return this.globalSilences.filter(silence => silence.status.state == 'active');
+        }
+    },
+    mounted: function () {
+        this.fetchAlerts();
+        this.fetchSilences();
+    }
+})
+
```
Asuforce: del line

promgen/proxy.py
```diff
@@ -174,3 +180,70 @@ def get(self, request):
...
+class ProxyDeleteSilence(View):
+    def delete(self, request, silence_id):
+        url = urljoin(
+            settings.PROMGEN["alertmanager"]["url"], "/api/v1/silence/%s" % silence_id
+        )
+        response = util.delete(url)
+        return HttpResponse(
+            response.text, status=response.status_code, content_type="application/json"
+        )
+
```
Asuforce: del line

promgen/templates/promgen/global_messages.html
```diff
@@ -0,0 +1,8 @@
...
+    <button type="button" class="close" data-dismiss="alert">
+        <span aria-hidden="true">&times;</span>
+    </button>
+    {{ message }}
+</div>
+{% endfor %}
```
Asuforce: add line

promgen/templates/promgen/project_detail_notifiers.html
```diff
@@ -19,8 +20,8 @@
     <a href="{% url 'project-notifier' project.id %}" class="btn btn-primary">{% trans "Register Notifier" %}</a>
     <form action="{% url 'project-notifier' project.id %}" style="display:inline" method="post">{% csrf_token %}
       <input type="hidden" name="sender" value="promgen.notification.user">
-      <input type="hidden" name="value" value="{{request.user.username}}"/>
+      <input type="hidden" name="value" value="{{request.user.username}}" />
       <button class="btn btn-primary">{% trans "Subscribe to Notifications" %}</button>
     </form>
   </div>
-</div>
+</div>
```
Asuforce: add line

promgen/templates/promgen/global_messages.html
kfdm: Ah, that part is actually still django template code it's unfortunate that the vue js and django templates look the same (so in some places I had to have django ignore vuejs blocks)

Asuforce: ok I understand.

PR #129, opened 2019-02-28
[BUGFIX] Fix computed labels

kfdm:
Removes some custom logic that was likely wrong and simplify using native javascript filter and map functions

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -60,17 +60,6 @@ var app = new Vue({
             }
             this.showSilenceForm(event);
         },
-        filterBy: function (source, label) {
-            let labels = new Set();
-            for (var a in source.filter(x => x.status.state == 'active')) {
-                for (var l in source[a].labels) {
-                    if (l == label) {
-                        labels.add(source[a].labels[label]);
-                    }
-                }
-            }
-            return labels;
-        },
         fetchSilences: function () {
             let this_ = this;
             fetch('/proxy/v1/silences')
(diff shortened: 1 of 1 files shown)


promgen/static/js/promgen.vue.js
```diff
@@ -103,23 +92,47 @@ var app = new Vue({
                 });
 
         }
-
     },
     computed: {
         alertLabelsService: function () {
-            return this.filterBy(this.globalAlerts, 'service');
+            return new Set(this.globalAlerts
```
Asuforce: I think better to define this process to a `method` and call it from computed property.
So maintenance will be easily.

kfdm: I previously wrote it as my own method, but I think that even though it's a few more lines in the computed properly itself, it's much more readable and more maintainable in this format

Asuforce: Like this

```js
  methods: {
    filterByService: function() {
      return new Set(
        this.globalAlerts
          .filter(x => x.status.state == 'active')
          .filter(x => x.labels.service)
          .map(x => x.labels.service)
          .sort()
      )
    },
    filterByProject: function() {
      return new Set(
        this.globalAlerts
          .filter(x => x.status.state == 'active')
          .filter(x => x.labels.project)
          .map(x => x.labels.project)
          .sort()
      )
    },
    filterByAlertName: function() {
      return new Set(
        this.globalAlerts
          .filter(x => x.status.state == 'active')
          .filter(x => x.labels.alertname)
          .map(x => x.labels.alertname)
          .sort()
      )
    }
  },
  computed: {
    alertLabelsService: function() {
      return this.filterByService
    },
    alertLabelsProject: function() {
      return this.filterByProject
    },
    alertLabelsRule: function() {
      return this.filterByAlertName
    },
    silenceLabelsService: function() {
      return this.filterByService
    },
    silenceLabelsProject: function() {
      return this.filterByProject
    },
}
```

Asuforce: I'm sorry. This code was wrong.

PR #133, opened 2019-03-06
[INTERNAL] Refactor Breadcrumb from template to simple_tag

kfdm:
Instead of doing a lot of logic in a template, refactor our logic into a simple_tag so that we can take advantages of iterators to simplify our logic and provide something easier to extend


kfdm:
Good catch! Should be fixed now :)

PR #172, opened 2019-08-17
Refactor apk commands in Dockerfile

PeterDaveHello:
Use `--no-cache` with `apk add` and remove `apk update` to prevent
leaving any local cache files in the immediate layers, which will make
the Docker image more slim/light.


kfdm:
Oh thank you. Good to know!

PR #185, opened 2019-10-10
Modify to the correct request method

shokada:
The request method of ```/api/v1/alerts```  is POST.
https://github.com/line/promgen/blob/master/promgen/views.py#L1034


kfdm:
Thanks!

PR #198, opened 2019-11-14
[INTERNAL] Refactored rule rendering to use Renderer and Serializer classes

kfdm:
Refactor to use rest_framework Serializer so that we can better consolidate our logic for rendering (and prefetching) a list of Rule objects and rendering them in the correct format


shokada:
Could you check Github Actions build failed and codecov failed?

PR #202, opened 2019-12-04
Change minimum coverage ratio.

shokada:
The codecov report is sometimes useful as a glance to see how much is being covered, but we do not want it to fail our pull request because of it so change minimum coverage ratio to 0%.
https://docs.codecov.io/docs/commit-status#section-target

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 0%
+    patch:
+      default:
+        target: 0%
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,8 @@
+coverage:
```
kfdm: Looks good! Lets add a comment in this file to explain why we set the targets to 0%

shokada: I Added a comment.

Issue #218, opened 2019-12-12
Rule label Expression should be in a code block.


sundy-li:
https://github.com/line/promgen/blob/master/promgen/templates/promgen/rule_detail.html#L71
https://github.com/line/promgen/blob/master/promgen/templates/promgen/rule_block.html#L14

If we put a expression like `{{$instance}}` in label value, the html will not render. So we suggest it in a    
code block.

```
-                    <li class="label label-primary">{{k}}:{{v}}</li>
+                    <li class="label label-primary">
+                       <code v-pre>{{k}}:{{v}}</pre></code>
+                    </li>
```

sundy-li:
I will create a pr.

PR #219, opened 2019-12-12
ISSUES #218, fix Rule label Expression render error

sundy-li:


.gitignore
```diff
@@ -13,6 +13,7 @@
 build
 coverage.xml
 dist
+.vscode
```
kfdm: Editor files should typically go in your personal `~/.gitignore`

promgen/templates/promgen/rule_block.html
```diff
@@ -11,7 +11,7 @@
     {% endif %}
     <ul>
       {% for k,v in rule.labels.items|dictsort:0 %}
-      <li class="label label-primary">{{k}}:{{v}}</li>
+      <li class="label label-primary"><code v-pre>{{k}}:{{v}}</pre></code></li>
```
kfdm: Can you give me a screenshot of where the problem is? I'm having difficulty understanding from your example and PR.

I didn't think it was normally possible to use `{{$instance}}` in label values (only in annotations) but I could be misunderstanding something

I believe there is also a syntax error here because you have an extra `</pre>`

It may be possible to handle the same thing with just the existing `<li>` tag doing something like `<li v-pre class="label label-primary">` but I would like a screenshot of where the error is actually happing to better understand the correct fix

sundy-li:
I am using labels   `instance  :  {{ $labels.instance}}`

The source html:
image

The error:
image

sundy-li:
We want the rule to have the instance labels, so we can alert the message with instances.

Prometheus will auto set `{{ $labels.instance}}` to be be `$instance` in alert messages.

kfdm:
Could I have a screenshot of the UI? In your example case, there should be no reason to add `instance` or `job` labels since they should automatically come from the alert data itself (or be empty)

sundy-li:
image

The labels are not standard and can be customized, just like exported_instance.

kfdm:
Typically the only place I have ever seen `{{ $templates }}` used are in annotations

sundy-li:
> Typically the only place I have ever seen `{{ $templates }}` used are in annotations

Oh, I did not notice that, so we could send the labels to the annotations instead of labels...

sundy-li:
Thanks for your time.

kfdm:
For example, this is an alert I use here

```
alert: ExporterDown
annotations:
  grafana: https://grafana.example.com/dashobard?orgId=1&var-datasource={{ $externalLabels.datasource | urlquery }}&var-service={{ $labels.service | urlquery }}&var-project={{ $labels.project | urlquery }}
  runbook: https://wiki.example.com/documentation
  summary: <!here|here> {{ $labels.instance }} of job {{ $labels.job }} has been down for more than 5 minute.
expr: up == 0
for: 5m
labels:
  severity: major
```

Any of the labels that are in the alert query (For example `up{service="foo", job="bar", instance="example.com:1234"}` will automatically be sent in the alert and do not need to be manually overridden.

sundy-li:
Got that, really thank you!

PR #242, opened 2020-02-25
Filter alerts

shokada:


promgen/tasks.py
```diff
@@ -12,6 +13,12 @@
 
 logger = logging.getLogger(__name__)
 
+@shared_task
+def index_alert(alert_pk):
+    alert = models.Alert.objects.get(pk=alert_pk)
+    labels = json.loads(alert.body)["commonLabels"]
```
kfdm: I think this could be written as `labels = alert.json.get('commonLabels')` since Alert model defines a `.json` property

shokada: Fixed

promgen/templates/promgen/alert_list.html
```diff
@@ -4,7 +4,12 @@
 
 {% breadcrumb label="Alerts" %}
 
-{% include 'promgen/pagination_short.html' %}
+<form class="navbar-form navbar-left" action="/alert">
+    <div class="form-group">
+        <input name="search" type="text" value="{{ search }}" class="form-control" placeholder="Service, Project">
```
kfdm: I think you can write this as `value="{{ request.GET.search }}"`  and then you do not have to implement `get_context_data` in the view

promgen/views.py
```diff
@@ -1019,13 +1019,28 @@ def post(self, request, *args, **kwargs):
...
         tasks.process_alert.delay(alert.pk)
+        tasks.index_alert.delay(alert.pk)
         return HttpResponse("OK", status=202)
 
 
 class AlertList(LoginRequiredMixin, ListView):
     paginate_by = 20
-    queryset = models.Alert.objects.order_by('-created')
 
+    def get_context_data(self, **kwargs):
```
kfdm: I think I would probably leave the queryset at the top, and then implement it something like 

```python
class AlertList(LoginRequiredMixin, ListView):
    paginate_by = 20
    queryset = models.Alert.objects.order_by("-created")

    def get_queryset(self):
        search = self.request.GET.get("search")
        if search:
            return self.queryset.filter(
                Q(alertlabel__name="Service", alertlabel__value__icontains=search)
                | Q(alertlabel__name="Project", alertlabel__value__icontains=search)
            )
        return self.queryset
```

kfdm:
May also be worthwhile to implement a management command [1] to allow to queue alerts to be re-indexed.

```python
for alert in Alert.objects.all():
    task. index_alert.delay(alert.pk)
```


[1]: https://docs.djangoproject.com/ja/2.2/howto/custom-management-commands/

promgen/views.py
shokada: fixed

promgen/templates/promgen/alert_list.html
shokada: fixed

shokada:
> May also be worthwhile to implement a management command 1 to allow to queue alerts to be re-indexed.
I will implement that.

shokada:
@kfdm How about this? https://github.com/line/promgen/pull/242/commits/9a1307e06f0a3a96bb478cf26af78ef1c17be5ca

promgen/views.py
```diff
@@ -1033,8 +1033,15 @@ def get_queryset(self):
             return self.queryset.filter(
                 Q(alertlabel__name="Service", alertlabel__value__contains=search)
                 | Q(alertlabel__name="Project", alertlabel__value__contains=search)
+                | Q(alertlabel__name="Job", alertlabel__value__contains=search)
             )
-        return self.queryset
+
+        qs = self.queryset
+        for key, value in self.request.GET.items():
+            if key == "page" or key == "search":
```
kfdm: Could probably write this as `if key in ["page", "search"]`

promgen/views.py
```diff
@@ -1033,8 +1033,15 @@ def get_queryset(self):
             return self.queryset.filter(
                 Q(alertlabel__name="Service", alertlabel__value__contains=search)
                 | Q(alertlabel__name="Project", alertlabel__value__contains=search)
+                | Q(alertlabel__name="Job", alertlabel__value__contains=search)
```
kfdm: May be better to do `alertlabel__value__icontains` so that upper/lower case do not matter

promgen/views.py
shokada: fixed

promgen/views.py
shokada: fixed

PR #249, opened 2020-02-28
Add dryrun mode

shokada:


promgen/management/commands/alerts-index.py
```diff
@@ -7,7 +7,21 @@
...
 
 class Command(BaseCommand):
-    def handle(self, **kargs):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dry-run',
+            dest='dryrun',
+            action='store_true',
+            help='show what would have been inserted',
+        )
```
kfdm:
```suggestion
        parser.add_argument(
            '--force',
            dest='dryrun',
            action='store_false',
            help='Defaults to dry run. Use to execute operation',
        )
```

In `alerts-prune` we use `--force` for our flag, so maybe it's best to match the same here.

promgen/management/commands/alerts-index.py
```diff
@@ -7,7 +7,21 @@
...
+
+    def handle(self, dryrun, **kargs):
         for alert in models.Alert.objects.filter(alertlabel__isnull=True):
+            if dryrun:
+                labels = alert.json.get("commonLabels")
+                for name, value in labels.items():
+                    self.stderr.write("alert_id: %s, name: %s, value: %s" % (alert.pk, name, value))
```
kfdm: Probably no need to loop through the labels here. Can write it as

```suggestion
                labels = alert.json.get("commonLabels")
                self.stderr.write("alert_id: %s, labels: %s" % (alert.pk, labels))
```

PR #250, opened 2020-03-02
Fix race condition

shokada:


promgen/tasks.py
```diff
@@ -43,6 +43,8 @@ def process_alert(alert_pk):
                 alert.delete()
                 return
 
+    tasks.index_alert.delay(alert.pk)
```
kfdm:
```suggestion
    index_alert.delay(alert.pk)
```

`index_alert` already exists in this file, so no need to reference the module

promgen/tasks.py
```diff
@@ -8,7 +8,7 @@
 from atomicwrites import atomic_write
 from celery import shared_task
 
-from promgen import models, prometheus, util, notification
+from promgen import models, prometheus, util, notification, tasks
```
kfdm: This isn't necessary and would cause an import loop

PR #258, opened 2020-03-16
Add toggle button for notifier

shokada:


promgen/views.py
```diff
@@ -350,6 +350,16 @@ def post(self, request, pk):
         return JsonResponse({'redirect': exporter.project.get_absolute_url()})
 
 
+class NotifierToggle(LoginRequiredMixin, View):
+    def post(self, request, pk):
+        sender = get_object_or_404(models.Sender, id=pk)
+        sender.enabled = not sender.enabled
+        sender.save()
+        signals.trigger_write_config.send(request)
```
kfdm: No need to trigger writing Prometheus' configuration since Notifiers are completely internal to Promgen

shokada: fixed

promgen/migrations/0018_sender_enabled.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.10 on 2020-03-14 03:10
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('promgen', '0017_alter_alertlabel_field'),
```
kfdm: Will need to rename this to `0019_sender_enabled.py` and fix this to 

```suggestion
        ('promgen', '0018_default_exporter_scheme'),
```

due to the earlier PR that was already merged 🙇

PR #262, opened 2020-03-30
Enable editing of exporter settings

shokada:


promgen/templates/promgen/exporter_form.html
```diff
@@ -11,7 +11,7 @@ <h1>Project: {{ project.name }}</h1>
 
 <div class="row">
 
-<form id="exporter_form" action="{% url 'project-exporter' project.id %}" method="post">{% csrf_token %}
+<form id="exporter_form" action="{{ url }}" method="post">{% csrf_token %}
```
kfdm: I think you can just remove `action="{{ url }}"` and let it default to POSTing to the same page. Then you don't need to do extra work in `get_context_data`

promgen/views.py
```diff
@@ -333,6 +333,19 @@ def post(self, request, pk):
...
+    model = models.Exporter
+    button_label = _("Exporter Update")
+    form_class = forms.ExporterForm
+
+    def get_success_url(self):
+        return reverse('project-detail', args=[self.object.project_id])
+
+    def get_context_data(self, **kwargs):
+        context = super(ExporterUpdate, self).get_context_data(**kwargs)
+        context["project"] = self.object.project
```
kfdm: I think you might be able to replace this using `get_initial`

```
def get_initial(self):
    return {"project": self.object.project}
```

But I need to finish testing myself

promgen/templates/promgen/exporter_form.html
shokada: `url` is also used in `ExporterRegister`, so I cannot remove `action="{{ url }}"`
https://github.com/line/promgen/pull/262/files#diff-4148d3344e097bea3c3c8f89325b0f57R629

kfdm: Yes, I think you can remove it from both :)

shokada: I'm sorry. I understood.
fixed https://github.com/line/promgen/pull/262/commits/d39efd4940a5f02f32f9c127a21e44ee3e6f0a76

promgen/views.py
shokada: ```ExporterRegister``` uses ```ProjectMixin``` to set ```project``` in the context.
https://github.com/line/promgen/blob/master/promgen/views.py#L608

Is it better to modify ```ExporterRegister``` to use ```get_initial``` as well?

kfdm: In that case, can probably use ProjectMixin the same way

shokada: Since the path of ExporterUpdate  is ```/exporter/<exporter-pk>/edit```, I don't think we can use ProjectMixin in ExporterUpdate

PR #264, opened 2020-03-31
Alter unique_together for exporter

shokada:


kfdm:
Looks like this was broken in #259 when the BaseExporter model was refactored.

promgen/models.py
```diff
@@ -349,7 +349,7 @@ class Exporter(BaseExporter):
 
     class Meta:
         ordering = ["job", "port"]
-        unique_together = (("job", "port", "path"),)
+        unique_together = (("job", "port", "path", "scheme", "project"),)
```
kfdm: I think the `unique_together` (and the migration) are the only changes required. What is the reason for the other changes ?

shokada: I set ```project``` in ```unique_together```, but the validation doesn't work because ```project``` is not set in the form.
So, I added ```get_form_kwargs``` and ```full_clean``` by referring to the following.
https://stackoverflow.com/a/32261039

PR #268, opened 2020-04-02
Select multiple servers to silence

shokada:


kfdm:
I'm not sure why CircleCI did not run, but since this is all javascript, I'll go ahead and merge it (I'm not sure how to re-trigger without commiting something)

PR #272, opened 2020-04-09
Allow to edit notifier settings

shokada:
The following error occurs when using `Form`, so I use `ModelForm`.
```
__init__() got an unexpected keyword argument 'instance'
```

### DIFF
--- a/promgen/templates/promgen/notifier_detail.html
+++ b/promgen/templates/promgen/notifier_detail.html
@@ -1,0 +1,27 @@
+{% extends "base.html" %}
+{% load promgen %}
+{% block content %}
+
+<div class="page-header">
+  <h1>Notifier</h1>
+</div>
+
+{% breadcrumb object 'Update Notifier' %}
+
+<div>
+    <div role="tabpanel" class="tab-pane panel panel-default">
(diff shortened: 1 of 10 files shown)


promgen/templates/promgen/notifier_detail.html
```diff
@@ -0,0 +1,27 @@
...
+{% breadcrumb object 'Update Notifier' %}
+
+<div>
+    <div role="tabpanel" class="tab-pane panel panel-default">
+        <form action="" method="post">{% csrf_token %}
+          <div class="panel-body">
+            <pre class="help-block">{{ driver.help|urlize }}</pre>
+          </div>
+          <table class="table">
+            {{ form.as_table }}
```
kfdm, line 18: We can not use form.as_table here, because it will expose `value` which is often a secret.  The form should only allow replacing the value, without exposing what the previous value was

shokada: I keep using form.as_table and set the value to empty in `get_initial`.
https://github.com/line/promgen/pull/272/commits/1e45a24be4f290f2a5188d6c8a615c083b4ecf3f

PR #283, opened 2020-06-04
docker-compose: fix the issue where celery broker was unable to push cfg

mtetreault:
Those changes should allow one to get started.

With those change, everything should work out of the box at the exception of the `admin/site/sites` thing that still needs to be configured.

The `links` have been removed because it's deprecated and is not required.
https://docs.docker.com/compose/compose-file/compose-file-v2/#links

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 # Reference for exteranl_links and bridge containers
 # https://blog.virtualzone.de/2016/09/docker-compose-link-containers-outside-compose-file-using-external_links.html
 
-version: '2'
+version: '2.1'
 services:
   base:
     build: .
@@ -14,30 +14,28 @@ services:
 
   web:
     extends: base
-    command: web -b 0.0.0.0:8000
+    command: web -b 0.0.0.0:8000 --reload
     ports:
       - "8080:8000"
-    links:
-      - mysql
-      - redis
     environment:
-      DEBUG: 1
+      - DEBUG=1
+      - ALLOWED_HOSTS=127.0.0.1,localhost,web
 
   worker:
     extends: base
-    command: worker -l info --queues localhost,celery
-    links:
-      - prometheus
-      - mysql
-      - redis
+    command: worker -l info --queues localhost,celery,prometheus,alertmanager
     volumes:
       - ./docker:/etc/prometheus
 
   prometheus:
     image: prom/prometheus
-    links:
-      - alertmanager
-      - blackbox
+    user: "root"
+    command:
+      - --config.file=/etc/prometheus/prometheus.yml
+      - --storage.tsdb.path=/prometheus
+      - --web.console.libraries=/usr/share/prometheus/console_libraries
+      - --web.console.templates=/usr/share/prometheus/consoles
+      - --web.enable-lifecycle
     ports:
       - "9090:9090"
     volumes:
@@ -47,14 +45,14 @@ services:
   blackbox:
     image: prom/blackbox-exporter
     network_mode: bridge
+    ports:
+      - "9115:9115"
     volumes:
       - ./docker:/etc/prometheus
 
   alertmanager:
     image: prom/alertmanager
     command: --config.file=/etc/prometheus/alertmanager.yml
-    links:
-      - web
     ports:
       - "9093:9093"
     volumes:


docker-compose.yml
```diff
@@ -14,30 +14,28 @@ services:
 
   web:
     extends: base
-    command: web -b 0.0.0.0:8000
+    command: web -b 0.0.0.0:8000 --reload
```
kfdm, line 17: In general `--reload` is really more for a development server so this should probably be removed.

docker-compose.yml
```diff
@@ -14,30 +14,28 @@ services:
...
+      - ALLOWED_HOSTS=127.0.0.1,localhost,web
 
   worker:
     extends: base
-    command: worker -l info --queues localhost,celery
-    links:
-      - prometheus
-      - mysql
-      - redis
+    command: worker -l info --queues localhost,celery,prometheus,alertmanager
```
kfdm, line 26: Typically the worker only needs to listen to 2 queues at max. 
`celery` is the default queue for almost everything (notifications, etc) and there should be a single named queue for each Prometheus worker that matches is host name. In the case of the docker container it might just be `prometheus` so perhaps this command should be `--queues prometheus,celery`

docker-compose.yml
```diff
@@ -14,30 +14,28 @@ services:
...
     volumes:
       - ./docker:/etc/prometheus
 
   prometheus:
     image: prom/prometheus
-    links:
-      - alertmanager
-      - blackbox
+    user: "root"
```
kfdm, line 32: In general, I think it's probably best to run as a non-root user

docker-compose.yml
```diff
@@ -14,30 +14,28 @@ services:
...
+    command:
+      - --config.file=/etc/prometheus/prometheus.yml
+      - --storage.tsdb.path=/prometheus
+      - --web.console.libraries=/usr/share/prometheus/console_libraries
+      - --web.console.templates=/usr/share/prometheus/consoles
+      - --web.enable-lifecycle
```
kfdm, line 38: I think all the defaults *should* be ok, so I'm unsure the reason to add them here.

docker-compose.yml
```diff
@@ -47,14 +45,14 @@ services:
   blackbox:
     image: prom/blackbox-exporter
     network_mode: bridge
+    ports:
+      - "9115:9115"
```
kfdm, line 49: blackbox_exporter is typically not accessed outside of the cluster, so normally the ports wouldn't need to be exported.

docker-compose.yml
mtetreault: You're right, this will be removed.

docker-compose.yml
mtetreault: I agree with you that it should be runned as a non-root user.

The original issue is that prometheus couldn't create the required files&folders.
```
prometheus_1    | level=error ts=2020-06-05T13:30:24.569Z caller=query_logger.go:87 component=activeQueryTracker msg="Error opening query log file" file=/prometheus/queries.active err="open /prometheus/queries.active: permission denied"
prometheus_1    | panic: Unable to create mmap-ed active query log
```
This is due to the fact that by default the folders were created with the following user:groups

``` bash
drwxr-xr-x 6 systemd-coredump root       4096 Jun  5 09:30 mysql_data
drwxr-xr-x 2 root             root       4096 Jun  5 09:30 prom_data
```
It looks like prometheus uses the nobody user https://github.com/prometheus/prometheus/blob/master/Dockerfile#L19

What do you think about using the user nobody as prometheus does?

The folder creation could then be done by an entrypoint script.

docker-compose.yml
mtetreault: The command option is overridden to add the `--web.enable-lifecycle` which is required to allow prometheus to reload its configuration. 

```
Prometheus can reload its configuration at runtime. If the new configuration is not well-formed, the changes will not be applied. A configuration reload is triggered by sending a SIGHUP to the Prometheus process or sending a HTTP POST request to the /-/reload endpoint (when the --web.enable-lifecycle flag is enabled). This will also reload any configured rule files
```

docker-compose.yml
mtetreault: I don't have a strong opinion on this point. I found it useful to have access to the page, mainly during testing, this can be removed.

docker-compose.yml
mtetreault: Alright, it seems to work properly with:

```
command: worker -l info --queus prometheus,celery
```

docker-compose.yml
kfdm: Using the same prometheus user (nobody) should be fine.

docker-compose.yml
kfdm: That's a good catch then. I suppose I need to test more carefully then because I never noticed this before (in production I do not deploy with docker so I've only tested the web part in docker)

kfdm:
Closing in favor of #402

PR #287, opened 2020-06-10
Fix collectstatic in the Dockerfile

nikita2206:
When collectstatic runs as `promgen` user it doesn't have access to dump files in `/root/.cache/promgen` (also not sure about the use of this directory)


kfdm:
Checking on the most recent master branch,
If I build a new container `docker-compose build base` 
and login to it `docker-compose run base /bin/sh`
It seems like the paths look correct `promgen diffsettings`
and the files show up as expected `find  ~/.cache/promgen -type f `

What type of error were you seeing ?

kfdm:
Closing due to lack of response

PR #288, opened 2020-06-11
[BUGFIXES] Use alternate Vue delimiters to fix double rendering bugs

kfdm:
Promgen often needs to deal with three different template languages, `Djagno`, `Vue`, and `Go`, and they all use the same `{{ }}` pattern for declaring templates. This shows up frequently on the Django side, where the `{% verbatim %}` needs to be used to ensure that **Vue** templates are not accidentally interpreted as **Django** templates (this happened a lot in initial development.

Recently, there have been cases when **Go** templates were rendered out to HTML from the rule editor, and then mistakenly rendered by **Vue**, causing the **Vue** renderer to throw an error and leave the page blank.

1. Fix one part of the **Django**, **Vue**, **Go** template mismatch, by changing the delimiters used so that they do not conflict. (This will also remove the requirement to use `{% verbatim %}` in several places)
2. There is still the possibility that a user could write a `[[ ]]` tag somewhere (intentionally for formatting reasons or by mistake) so need to go through and find places where this user input needs to be handled via a **Vue** `v-pre` tag. (Some of this may also be done in future pull requests as they're found)

Remaining tasks
- [ ] Finish some more auditing of `v-pre` tags
- [ ] Add developer note about changed **vue** delimiters

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -15,6 +15,7 @@ var dataStore = {
 
 var app = new Vue({
     el: '#vue',
+    delimiters: ['[[', ']]'],
     data: dataStore,
     methods: {
         toggleTarget: function (target) {
--- a/promgen/templates/base.html
+++ b/promgen/templates/base.html
@@ -39,28 +39,21 @@
   <script src="{% static "/js/promgen.vue.js" %}"></script>
   {% block javascript %}{% endblock %}
 
-
   <datalist style="display:none" id="common.labels">
-    <select>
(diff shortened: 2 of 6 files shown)


promgen/static/js/promgen.vue.js
```diff
@@ -15,6 +15,7 @@ var dataStore = {
 
 var app = new Vue({
     el: '#vue',
+    delimiters: ['[[', ']]'],
```
Asuforce, line 18: 👍

PR #299, opened 2020-12-02
Update install docs to mention Makefile

kfdm:
In the near future, the `bootstrap` command still needs to likely be split up better, but this should make progress on a bit easier install

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ make migrate
 # Run tests
 make test
 # Run development server
-make runserver
+make run
 ```
 
 For development and testing, it may be useful to make a symlink for running Promgen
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -2,32 +2,45 @@
...
 
 # Setup or update database schemas, create initial user and shard
-promgen migrate
+make migrate
 # Run tests
-promgen test
+make test
 # Run development server
-promgen runserver
+make runserver
```
sh-kawakami, line 36: I think, right command is `make run`?

sh-kawakami (changes requested):

CONTRIBUTING.md
kfdm: Ah yes. Good catch

sh-kawakami (approved):

CONTRIBUTING.md
sh-kawakami: It's fine, thank you!

Issue #302, opened 2020-12-03
run docker run --rm -it -v ~/.config/promgen:/etc/promgen/ line/promgen bootstrap  the system recommend  1146, "Table 'prome.promgen_shard' doesn't exist"


liublcom:
hi:
i have run the command docker run --rm -it -v ~/.config/promgen:/etc/promgen/ line/promgen bootstrap,
then i put the mysql driver mysql://prome:Prome2020@10.0.184.106/prome when the system recommend to put DATABASE_URL,the last the system have this error :Running django system checks
Running django system checks
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/base.py", line 71, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 209, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 315, in _query
    db.query(q)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py", line 239, in query
    _mysql.connection.query(self, query)
MySQLdb._exceptions.ProgrammingError: (1146, "Table 'prome.promgen_shard' doesn't exist")

i found the database prome  don't have any table 。

i have do as followers
mkdir -p ~/.config/promgen
chmod 777 ~/.config/promgen
docker run --rm -it -v ~/.config/promgen:/etc/promgen/ line/promgen bootstrap

kfdm:
Hi @liublcom 
I think we can keep it all in a single issue, so probably no need to have #301 and #300 

As you have noticed (and I have also conincidentally noticed this week) there's a bit of a weird interaction in the `promgen bootstrap` command. The first time fails because it has some of the django defaults cached, but then it passes the second time. I'm working with my team to decide the best way to handle it, but it may involve splitting the command into two parts. I'm sorry for the confusion 🙇

liublcom:
@kfdm ：
      #301 and #300  is not need to care.
     MySQLdb._exceptions.ProgrammingError: (1146, "Table 'prome.promgen_shard' doesn't exist") ,this message tell me we should first to init the database.

liublcom:
@kfdm 
         excuse ，Are you following this issue。

kfdm:
[...]

PR #303, opened 2020-12-03
Split bootstrap checks

kfdm:
The bootstrap command has an issue where it is difficult to both configure the settings and test at the same time, because there is no easy way to also reload the django settings. This means we need to separate the commands. We also want to run some tests that require the database, but we can't have them enabled by default or other commands like migrate will be stuck in a chicken/egg problem. To get around this we create a decorator to check for database access for our checks that require it.

This is a continuation of documentation fixes for #299 and should handle the issue in #302


sh-kawakami (approved):

sh-kawakami:
looks good!

PR #336, opened 2021-04-07
Closes #335: Notify Slack using proxy

### LINKED ISSUE #335: Add feature to allow slack notification to use proxy
It would be nice to have a feature to notify Slack using a proxy server without modifying the proxy configuration of the whole application.

monicajss:
Solution of issue #335.


promgen/notification/slack.py
```diff
@@ -37,6 +37,14 @@ class NotificationSlack(NotificationBase):
...
 
     def _send(self, url, data):
+        
+        proxy = self.config('proxies')
+        if proxy is not None and proxy != "":
+            proxies = {
+                'http': proxy,
+                'https': proxy,
+            }
+
```
kfdm, line 47:
```suggestion
        kwargs = {}
        proxy = self.config('proxies')
        if proxy is not None and proxy != "":
            kwargs['proxies'] = {
                'http': proxy,
                'https': proxy,
            }
        
```

promgen/notification/slack.py
```diff
@@ -45,5 +53,11 @@ def _send(self, url, data):
...
-
-        util.post(url, json=json).raise_for_status()
+         
+        if proxy is not None and proxy != "":
+            logger.info('Notifying Slack using proxy')
+            util.post(url, json=json, proxies=proxies).raise_for_status()
+        else:
+            logger.info('Notifying Slack without proxy')
+            util.post(url, json=json).raise_for_status()
+
```
kfdm, line 63:
```suggestion
         
        util.post(url, json=json, **kwargs).raise_for_status()
```

kfdm (changes requested):
Looks good! I think one change I would typically do, is for some of these optional parameters, have an `extra` or `kwargs` (keyword argument) dictionary, to contain the extra parameters such as `proxies`. Then you can call `util.post(url, json=json, **kwargs).raise_for_status()` without needing an additional `if/else` conditional check

promgen/notification/slack.py
monicajss: Thanks for the advice, actually it looks way better using the kwargs, we will update the code, test it and update the pull request, thanks a lot.

kfdm (approved):
Looks good!

PR #348, opened 2021-05-26
Notification handling improvements

kfdm:
- Be more explicit about handling defaults for NotificationBase.config and util.setting
- Check AlertError in our unit test results
- Split send_alert into send_notification to better handle our counting logic

### DIFF
--- a/promgen/notification/__init__.py
+++ b/promgen/notification/__init__.py
@@ -50,7 +50,7 @@ class NotificationBase(object):
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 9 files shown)


sh-kawakami (approved):

m-tkg (approved):

kfdm:
- Be more explicit about handling defaults for NotificationBase.config and util.setting
- Check AlertError in our unit test results
- Split send_alert into send_notification to better handle our counting logic

PR #358, opened 2021-07-02
add databse migration command in README.md

kackey0-1:
I think it's better to have migration command to setup test environment.
Otherwise developer might execute  `docker-compose run web createsuperuser` 
before executing `docker-compose run web migrate`


kfdm (approved):

kfdm:
Thank you for the documentation update 😄

PR #372, opened 2021-11-13
[IMPROVEMENT] Improve performance of rendering alerts and silences

KarboniteKream:
## Description
This PR improves performance of rendering alerts and silences in Vue.

Currently, all alerts and silences are always rendered and hidden using the `collapse` CSS rule. This causes rendering delays on page load and whenever components need to be re-rendered (for example, on form input change; #370).

With changes introduced in this PR, all alerts and silences are now hidden behind `v-if`, so the HTML elements are not generated/rendered until the expression is truthy.

## Detailed changes
* Introduce `toggleComponent()`, which dynamically toggles the state of components
  * The initial state of all components is falsy, so they're not shown by default
  * Since the `components` object does not have any properties, it needs to be updated using `this.$set()`
* Use `toggleComponent()` for all elements that previously used `toggleTarget()` and `collapse`
  * Add `v-cloak` to top-level element of specific components
* Replace `v-show` with `v-if` so we skip rendering of `v-for` if it's not necessary
* Remove unnecessary `prevent` modifier on `@click`
* Improve caching of `computed` methods
  * We can reuse `filterActive{Alerts,Silences}`
  * Since we return a `Set`, there's no need to sort the values
* Rename `toggleTarget()` to `toggleCollapse()`
  * Alerts and silences on the project detail page are expanded by default, so `toggleComponent()` cannot be used

## Improvements
Before the improvements, the page would be mostly unresponsive until the `Alerts` block is rendered.
Note: All measurements are approximate page reload times without cache, and were performed with 1660 alerts, 25 silences and 6000 farms.

| Component | Before | After |
| ---- | ------ | ----- |
| Home page | 2s | 0.8s |
| Global alerts section | 2s | 2.5s |
| Services page | 5.2s | 1.9s |
| Link Farm page | 16s | 6.7s |

* Page loads are now completely responsive, except for `Link Farm` since that needs to render the HTML of 6000 farms.
* The global `Alerts` section, now takes 500ms longer to open, since the HTML elements need to be generated.
* Form input lag is almost nonexistent, unless the global `Alerts` section is open.
* The bottleneck now seem seem to be API responses

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -6,6 +6,7 @@
 Vue.config.devtools = true
 
 var dataStore = {
+    components: {},
     selectedHosts: [],
     newSilence: { 'labels': {} },
     globalSilences: [],
@@ -18,7 +19,11 @@ var app = new Vue({
     delimiters: ['[[', ']]'],
     data: dataStore,
     methods: {
-        toggleTarget: function (target) {
+        toggleComponent: function (component) {
+            let state = Boolean(this.components[component]);
+            this.$set(this.components, component, !state);
+        },
+        toggleCollapse: function (target) {
             let tgt = document.getElementById(target);
             tgt.classList.toggle('collapse');
         },
@@ -52,7 +57,7 @@ var app = new Vue({
             this.$delete(this.newSilence.labels, label)
         },
         showSilenceForm: function (event) {
-            document.getElementById('silence-form').classList.remove('collapse');
+            this.$set(this.components, 'silence-form', true);
             scroll(0, 0);
         },
         silenceAppendLabel: function (event) {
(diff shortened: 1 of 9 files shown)


promgen/templates/promgen/project_detail.html
```diff
@@ -40,9 +40,9 @@ <h1>
...
 </div>
 
-<div id="silence-project-{{ project.name|slugify }}" class="panel panel-warning" v-cloak v-show="silenceLabelsProject.has('{{project.name}}')">
+<div id="silence-project-{{ project.name|slugify }}" class="panel panel-warning" v-cloak v-if="silenceLabelsProject.has('{{project.name}}')">
   <div class="panel-heading">
-    <a @click.prevent="toggleTarget('silences-project-{{project.name|slugify}}')" class="btn btn-warning btn-sm" role="button">Silences</a>
+    <a @click="toggleCollapse('silences-project-{{project.name|slugify}}')" class="btn btn-warning btn-sm" role="button">Silences</a>
   </div>
   <table id="silences-project-{{project.name|slugify}}" class="table table-bordered table-condensed">
     <tr v-for="(silence, index) in filterActiveSilences" v-if="silence.labels.project == '{{project.name}}'">
```
KarboniteKream, line 48: Using `v-for` with `v-if` is [not recommended](https://v3.vuejs.org/style-guide/#avoid-v-if-with-v-for-essential). Additionally, this iterates over all active alerts/silences, which takes time, but currently doesn't look like a problem. I'll improve this in the next PR.

KarboniteKream (commented):
For the `Link Farm` page, we could improve the performance by limiting the number of items shown. Creating a separate API and controlling rendering in Vue would be one of the options.

promgen/static/js/promgen.vue.js
```diff
@@ -123,43 +128,33 @@ var app = new Vue({
...
+            return new Set(this.filterActiveSilences
                 .filter(x => x.labels.service)
                 .map(x => x.labels.service)
-                .sort()
             );
         },
         silenceLabelsProject: function () {
-            return new Set(this.globalSilences
-                .filter(x => x.status.state == 'active')
+            return new Set(this.filterActiveSilences
```
KarboniteKream, line 155: The behavior here has changed slightly. I saw that `filterActiveSilences` was changed from `state == 'active'` to `state != 'expired'`. I think we want the same behavior here?

kfdm (approved):

PR #373, opened 2021-11-20
[IMPROVEMENT] Improve list rendering in Vue

KarboniteKream:
## Description
This PR refactors the way alert and silence lists are rendered, by removing `v-if` from `v-for` directives. This way we avoid any runtime errors and unnecessary iterations.

## Detailed changes
* Update `computed` field to return `Map` instead of `Set`
* Use these new maps in `v-for` to avoid the need for additional filters
* Remove unused `index` variable in iterations

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -1,11 +1,11 @@
 /*
-# Copyright (c) 2019 LINE Corporation
-# These sources are released under the terms of the MIT license: see LICENSE
-*/
+ * Copyright (c) 2021 LINE Corporation
+ * These sources are released under the terms of the MIT license: see LICENSE
+ */
 
 Vue.config.devtools = true
 
-var dataStore = {
+const dataStore = {
     components: {},
     selectedHosts: [],
     newSilence: { 'labels': {} },
@@ -14,7 +14,7 @@ var dataStore = {
     globalMessages: []
 };
 
-var app = new Vue({
+const app = new Vue({
     el: '#vue',
     delimiters: ['[[', ']]'],
     data: dataStore,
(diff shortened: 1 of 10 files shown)


promgen/static/js/promgen.vue.js
```diff
@@ -92,27 +92,27 @@ var app = new Vue({
...
-                            silence.labels[matcher.name] = matcher.value
+                    // to make other code easier
+                    for (let silence of silences) {
+                        silence.labels = {};
+                        for (let matcher of silence.matchers) {
+                            silence.labels[matcher.name] = matcher.value;
                         }
                     }
+
+                    this_.globalSilences = silences;
```
KarboniteKream, line 107: Previous logic dynamically added properties to object observed by Vue, which can cause reactivity issues. This is solved by using an intermediate variable before assigning to `globalSilences`.

promgen/templates/promgen/project_detail.html
```diff
@@ -29,23 +29,23 @@ <h1>
...
 {% include "promgen/project_detail_configuration.html" %}
 
-<div class="panel panel-danger" v-cloak v-if="alertLabelsProject.has('{{project.name}}')">
+<div class="panel panel-danger" v-cloak v-if="activeProjectAlerts.has('{{project.name}}')">
   <div class="panel-heading">
     <a @click="toggleCollapse('alerts-project-{{project.name|slugify}}')" class="btn btn-danger btn-sm" role="button">Alerts</a>
   </div>
   <table id="alerts-project-{{project.name|slugify}}" class="table table-bordered table-condensed">
-    <tr v-for="(alert, index) in globalAlerts" v-if="alert.labels.project == '{{project.name}}'">
+    <tr v-for="alert in activeProjectAlerts.get('{{project.name}}')">
```
KarboniteKream, line 37: The behavior changed here (ditto for other places in this PR). Since the previous code was iterating over `globalAlerts`, it would also display inactive (suppressed) alerts, but only if there's any other active alerts (due to `alertLabelsProject.has()`).

The new logic only displays active alerts now. Is this change acceptable?

kfdm: Good catch on the bug. The global alerts should probably only show alerts set to active (not silenced) though the project/service specific ones should probably show all of them, with an extra indicator if it is silenced or not.

kfdm: I think this will work as is, and we can think about if it needs to be changed later 😄

kfdm (approved):

promgen/templates/promgen/project_detail.html
KarboniteKream: Sure, I can work on this after I finish with my currently planned PRs

PR #377, opened 2022-01-12
Move silence form to a separate component

KarboniteKream:
## Description
This PR moves the silence form to a separate component to prepare for future rework.

## Detailed changes
* Extract silence form into a new `<silence-form>` component
  * Template is now included as a X-Template ([reference](https://vuejs.org/v2/guide/components-edge-cases.html#X-Templates)). Without transpilation, we don't have access to single file components, and this way we don't have to keep it as a string
* Extract shared state into a simple state store ([reference](https://vuejs.org/v2/guide/state-management.html))
  * This allowed to simplify silence label update functions
* Cleanup `silence_form.html`
  * CSRF tokens are no longer needed, since we pass them with cookies
  * `.lazy` modifier is no longer necessary after #372 
* Use arrow functions to avoid `this` reassignment


promgen/static/js/promgen.vue.js
```diff
@@ -27,72 +51,33 @@ const app = new Vue({
...
-            this.$set(this.newSilence, 'labels', {});
-            this.$set(this.newSilence.labels, "instance", this.selectedHosts.join("|"));
-            for (key in event.target.dataset) {
-                this.$set(this.newSilence.labels, key, event.target.dataset[key]);
+        setSilenceLabels(labels) {
+            if (labels.target) {
+                silenceStore.setLabels(labels.target.dataset);
+            } else {
+                silenceStore.setLabels(labels);
             }
```
KarboniteKream, line 63: Another option to do this would be just pass `$event.target.dataset` at the call sites. WDYT?

kfdm: Yeah, that's one of the few places I'm unsure about. Could also have two methods

```javascript
        setSilenceLabels(labels) {
            silenceStore.setLabels(labels);
            silenceStore.showForm();
            scroll(0, 0);
        },
        setSilenceDataset(ele) {
            this.setSilenceLabels(ele.target.dataset);
        },
```

That might be cleaner

kfdm (approved):
Looks good! Makes some good progress in sorting out some of my earlier Vuejs mess 😆

Issue #346, opened 2021-05-17
[QUESTION] Prometheus Parameter


vabene1111:
I have a multi target exporter which is controlled via the `params` config directive. 

I could not find a way to pass my parameters to the target using promgen. Do i just manually add them to the URL/metrics path ?

I am relatively new to prometheus and still trying to understand promgen as well so i hope i haven't missed anything. 

example from 
https://prometheus.io/docs/guides/multi-target-exporter/#configuring-modules

```
- job_name: blackbox-http # To get metrics about the exporter’s targets
  metrics_path: /probe
  params:
    module: [http_2xx]
    target: [prometheus.io]
  static_configs:
    - targets:
      - localhost:9115
```

thanks for the help

vabene1111:
i tried adding it to the metrics path directly but that gets encoded by prometheus removing the `?` as described [here](https://stackoverflow.com/questions/40172415/question-mark-in-prometheus-metrics-path-gets-encoded)

kfdm:
Currently Promgen does not support the path parameter. It's something I would like to support, but recently I have been occupied with other tasks 🙇

vabene1111:
ok thanks for the information 👍

Issue #180, opened 2019-09-10
Add addition params feature into import prometheus configuration


tuan-phan:
Hi Team,
Currently, promgen configuration only support some params. I have some configuration need migrate to Promgen like:

> scrape_configs:
  - job_name: 'test_application'
    metrics_path: 'prometheus'
    params:
      token: ['xxxxxx']
    file_sd_configs:
    - files:
      - file-sd/jennifer_node*.json

jennifer_node*.json have this content
>   {
    "targets": [
      "http://atrkadas-stream1.travel-dev.dev.jp.local/healthcheck.html",
      "http://atrsrm-api.travel-dev.dev.jp.local/healthcheck.html"
    ],
    "labels": {
      "type": "internal",
      "platform": "cloud"
    }

Can we add some params like token, type, platform .... ?
It should be useful.
Thank you,

kfdm:
It's not possible now, but I have thought about supporting custom labels and params for scrape targets. Promgen would likely still pass its default labels (job/service/project) though you could handle that with relabel config

In your idea, how would you expect to handle the UI side of things?
Would the same token/custom label apply to all targets in a project, or would it apply only to a specific target

tuan-phan:
Hi @kfdm ,
Some case we need apply for whole project, some case we apply only to a specific target...
Example: 
The token will apply for all targets. But some labels like: app_team_name, host_group will be use for specific target
  {
    "targets": [
      "dev-mon-test2.dev.local:7900"
    ],
    "labels": {
      "app_team_name": "booking",
      "host_group": "dev-mon-test2",
      "platform": "private"
    }
  },
  {
    "targets": [
      "dev-mon-test1.dev.local:7900"
    ],
    "labels": {
      "app_team_name": "apm",
      "host_group": "dev-mon-test1",
      "platform": "cloud1"
    }
  },

dykow:
Hi!
Is it possible to do something like this now?
`targets.json`
```
[
    {
      "targets": [
        "x.x.x.x"
      ],
      "labels": {
        "job": "vmware"
      },
      "params": {
          "section": ["origin"]
      }
    }
]  
```
@kfdm

kfdm:
Currently there is no support for something like that 🙇

dykow:
> Currently there is no support for something like that bow

Oooh, ok. Thanks for the reply :)

PR #381, opened 2022-01-22
[IMPROVEMENT] Added exporter labels.

ekhvalov:
To resolve issues #346 and partially #180


promgen/templates/promgen/exporter_form.html
```diff
@@ -34,6 +34,40 @@ <h1>Project: {{ project.name }}</h1>
         <table class="table">
           {{ form.as_table }}
         </table>
+    <div class="labels-formset" style="padding: 8px">
+        {{ labels_formset.management_form }}
+        <div class="panel panel-default">
+            <div class="panel-heading" role="tab" id="labels-formset-panel-header">
+                <a role="button" data-toggle="collapse" href="#labels-formset-panel" aria-expanded="false"
```
kfdm, line 41: label form
Thinking back and forth a bit about this. It's probably good to have it toggle visible like you have it, but I'm wondering if we may need to make small adjustments to the text descriptions

promgen/templates/promgen/project_detail_exporters.html
```diff
@@ -10,7 +10,9 @@
       <th colspan="4">Actions</th>
     </tr>
   {% for exporter in project.exporter_set.all %}
-    <tr {% if not exporter.enabled %}class="promgen-disabled"{% endif %}>
+    <tr {% if not exporter.enabled %}class="promgen-disabled"{% endif %}
```
kfdm, line 13: mouseover labels

Even though we use the `title=` attribute in a few places to show extra information, I think in this case, additional labels are quite important, and we need to figure out how to make them always visible. Part of the challenge though is we do not currently have a lot of space there, so still thinking if we need to have a toggle collapse menu for it or not.

promgen/tests/test_prometheus.py
```diff
@@ -0,0 +1,273 @@
...
+from promgen import models
+from promgen import prometheus
+
+
+class PrometheusConfigTest(PromgenTest):
+    @mock.patch('django.dispatch.dispatcher.Signal.send')
+    def setUp(self, mock_signal):
+        user_id = 999
+        user = self.add_force_login(id=user_id, username='Tester')
+        self.__create_services([
```
kfdm, line 13: Instead of having all this structure in code, we should really be using fixtures for the tests. (Older, existing tests should also be migrated to fixtures  but that's something I should clean up)

https://docs.djangoproject.com/en/4.0/howto/initial-data/

kfdm (changes requested):
Overall, a good start! You covered a lot of the basic functionality and even added tests!!!

Part of the challenge for this (and Promgen overall) is how to make it easy, to modify a complicated configuration. There are lots of edge cases to consider. One that I quickly noticed, is that while you call out `__param_` labels (good!) they're not actually taken into account when using the Exporter text button, which could be quite confusing when Promgen and Prometheus scrape things differently.
test param missing

That combined with the other comment about displaying the custom labels (and not just using `title=` attribute, I think we still need to think through the exact workflow a little more.

I think my next two days will be a little busy, but hopefully that gives you some good feedback. I'll try to circle back to this PR in a few days.

promgen/templates/promgen/exporter_form.html
ekhvalov: What adjustments you want to make?

promgen/tests/test_prometheus.py
ekhvalov: OK, I'll use fixtures.

promgen/templates/promgen/project_detail_exporters.html
ekhvalov: Will button `Labels` be OK? Button

promgen/templates/promgen/exporter_form.html
ekhvalov: The main issue for me is the scrape exporter code. First of all there is a [TODO](https://github.com/line/promgen/blob/master/promgen/views.py#L620) in the code.  Probably API `/rest/project/<slug>/scrape` shoult know nothing about exporter register form fields and should receive a POST with body like `{"host":"blackbox", "port":9115, "scheme": "https", "path": "/probe", "query": {"target": "prometheus.io"}}`. If so, then `ExporterScrape` class could build the POST body from register exporter form then query API and return API response. Is that OK?
Maybe it is better to add a field `is_parameter` to `ExporterLabel`? In that case user is no need any more to use a prefix `__param_` and all logic about prefix could be moved to prometheus `render_config` function.
Also I have a question about API `/rest/project/<slug>/scrape`. Why exporter's id is omitted? The url could be `/rest/project/<slug>/exporter/<slug>/scrape`. In that case all url logic could be encapsulated in an `Exporter` itself.

promgen/tests/test_prometheus.py
ekhvalov: Done

promgen/templates/promgen/project_detail_exporters.html
ekhvalov: Button with dropdown table added.

ekhvalov:
New exporter scraping is fixed but scraping for existed exporters is still not working. We have to decide how to deal with it. I think  we have two different types of scraping, one for existed exporters and another for new ones. Probably we have to scrape new exporters using my solution and create API especially for existed exporters.
Also I have some questions.
If `is_parameter` field was added, is it correct to allow users to name labels with double underscore prefix?
Any validation is needed?

promgen/forms.py
```diff
@@ -93,6 +93,57 @@ class Meta:
...
 
+class ExporterLabelForm(ExporterForm):
+    class Meta:
+        model = models.ExporterLabel
+        exclude = ['exporter']
+
+        widgets = {
+            'name': forms.TextInput(attrs={'class': 'form-control', 'placeholder': 'Name'}),
+            'value': forms.TextInput(attrs={'class': 'form-control', 'placeholder': 'Value'}),
+            'is_parameter': forms.CheckboxInput()
```
kfdm, line 104: I think we **don't** want to use a parameter here (and then later build a string). I think having this as a toggle, and then try to generate strings from this later is confusing. If a user inputs a label as `foo` or `__param_foo` or `__invalid_foo` then that should be the exact name they see, and the exact name that is written into the Prometheus target.
We **do** however need to make sure we pass along the parameter when someone clicks the test version so we'll likely need to filter on that.

```python
# psudo code
params = {}
for key in labels:
  if key.startswith('__param'):
     params[ key[:7] ] = labels[key]
```

I think this will cause the least amount of confusion in the long run.

promgen/forms.py
```diff
@@ -93,6 +93,57 @@ class Meta:
...
+            return
+        labels_names = {}
+        parameters_names = {}
+        for form in self.forms:
+            name = form.cleaned_data.get('name')
+            if not name:
+                continue
+            if form.cleaned_data.get('is_parameter'):
+                if name in parameters_names:
+                    e = ValidationError('Duplicated name', code='invalid')
```
kfdm, line 121: I think I need to try to test this on my own, but something seems a bit off at first glance. I *think* all the checks for duplicated can probably just be left to the database and you shouldn't need custom logic in here.

promgen/models.py
```diff
@@ -370,6 +370,24 @@ def __str__(self):
...
 
+class ExporterLabel(models.Model):
+    name = models.CharField(
+        blank=False,
+        max_length=128,
+    )
+    value = models.CharField(
+        max_length=128,
+    )
+    is_parameter = models.BooleanField(default=False)
```
kfdm, line 381: as mentioned above, I think having a toggle for this is going to be more confusing to the user, than doing a bit of filtering on `key.startswith('__param')`

promgen/prometheus.py
```diff
@@ -132,12 +132,21 @@ def render_config(service=None, project=None):
         if exporter.path:
             labels["__metrics_path__"] = exporter.path
 
+        for label in exporter.exporterlabel_set.all():
+            if label.name not in labels:
+                label_name = '__param_{}'.format(label.name) if label.is_parameter else label.name
```
kfdm, line 137: As mentioned above, I think we don't want to use `is_parameter` but filter on `__param_` instead.

promgen/templates/promgen/exporter_form.html
kfdm: The reason the scrape code is as is, is because you can use the `test` button before it has been saved to the database, so we need to be able to POST the data to a test input when we don't already have an exporter_id.

promgen/tests/test_prometheus.py
```diff
@@ -0,0 +1,141 @@
...
+class PrometheusConfigTest(PromgenTest):
+    fixtures = ['testcases.yaml', 'testcases-prometheus.yaml']
+
+    @mock.patch('django.dispatch.dispatcher.Signal.send')
+    def setUp(self, mock_signal):
+        self.force_login(username='demo')
+
+    def test_create_config(self):
+        config = prometheus.create_config()
+        self.assertEqual(config, [
```
kfdm, line 17: Thanks!! Moving most of it to a filter really helps clean up the code. I think this config though can also be cleaned up a bit. You should be able to use our help function to put these test configurations as json as well

`TEST_CONFIG = tests.Data('examples', 'test-targets.json').json()`

or something similar

kfdm (changes requested):
Moving some of the data to fixtures helps clean up the tests a lot. This is great!
I don't think we want to depend on a `is_parameter` though, since I think that will add a good deal of confusion.

I did a quick pass today, but I will likely have to wait until next week to do a more detailed pass. I have a few tasks that will likely take the rest of my attention this week. Thank you for your patience 🙇

ekhvalov:
I have removed field 'is_parameter' and moved test config data to separate file in the `examples` folder.
There is one more thing I have to do, is to solve saved exporters testing. I need some time to do it, because I have to get familiar with rest-framework.

ekhvalov:
Exporters scraping is working via new REST endpoint. The old endpoint is still exists and could be removed if new one is OK.

kfdm:
I apologize for the long delay without comment. I have not forgotten this PR, I have just been preoccupied with other work tasks which have made it more difficult to prioritize this.

Advanced users may have specific, legitimate use cases for this, there is also a large set of less advanced users who will not understand the implications and are very likely to make things more complicated by adding additional custom labels. I'm somewhat reluctant to merge as is, again not because the idea or code are bad, but because we need to be very careful that we can clearly communicate to the user when this feature should be used (since it likely should not be used in the most common cases). Unfortunately, this April may also be a little bit scattered in terms of ability so I can not currently commit to when I'll be able to think through all the implications of this change. 🙇

PR #389, opened 2022-05-26
[BUGFIX] Fix for duplicate user subscribe

kfdm:
When using get_or_create, we want to ensure an owner, but we do not care
if someone else has ownership. Move our owner field to defaults to avoid a
duplicate sender in some cases.


vincent-olivert-riera (approved):
LGTM

PR #391, opened 2022-05-27
Enable copied rules by default

vincent-olivert-riera:
When overwriting site rules, a copy of the original rule is created. If a user
does that, it's because they prefer the copy (with their modifications) than the
original. Because of that it makes sense to enable the copy by default.

### DIFF
--- a/promgen/models.py
+++ b/promgen/models.py
@@ -473,7 +473,9 @@ class Rule(models.Model):
             self.name = '{}_{}'.format(self.name, slugify(content_object.name)).replace('-', '_')
             self.content_type = content_type
             self.object_id = object_id
-            self.enabled = False
+            # Enable the copy by default since it's more likely the user prefers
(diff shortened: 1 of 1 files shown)


promgen/models.py
```diff
@@ -473,7 +473,7 @@ def copy_to(self, content_type, object_id):
             self.name = '{}_{}'.format(self.name, slugify(content_object.name)).replace('-', '_')
             self.content_type = content_type
             self.object_id = object_id
-            self.enabled = False
+            self.enabled = True
```
kfdm, line 476: The original choice of `False` was probably arbitrarily chosen, but let us go ahead and add a comment either here, or as part of the function header to explain our default of True.

vincent-olivert-riera: Done in 0c18e28.

kfdm (approved):

PR #393, opened 2022-05-31
[IMPROVEMENT] Add support for SENTRY_SEND_DEFAULT_PII flag

kfdm:


promgen/settings.py
```diff
@@ -187,7 +187,13 @@
     from sentry_sdk.integrations.celery import CeleryIntegration
 
     os.environ.setdefault("SENTRY_RELEASE", __version__)
-    sentry_sdk.init(integrations=[DjangoIntegration(), CeleryIntegration()])
+    # By default we want to keep sentry's default of NOT sending user
+    # information, but in some environments, it's useful for debugging
+    # purposes so lets add a flag here that we can use.
+    sentry_sdk.init(
+        integrations=[DjangoIntegration(), CeleryIntegration()],
+        send_default_pii=env.bool("SENTRY_DEFAULT_PII", False),
```
vincent-olivert-riera, line 195: Like we do for other [Sentry common options](https://docs.sentry.io/platforms/python/configuration/options/#common-options) like [dsn](https://github.com/line/promgen/blob/master/promgen/settings.py#L184) or [release](https://github.com/line/promgen/blob/master/promgen/settings.py#L189), I think we should also follow the `SENTRY_<option name>` pattern.

So, instead of `SENTRY_DEFAULT_PII` I would prefer `SENTRY_SEND_DEFAULT_PII`.

This should also be reflected in the commit title:

`Add support for SENTRY_SEND_DEFAULT_PII flag`

vincent-olivert-riera (changes requested):

vincent-olivert-riera (approved):
LGTM

PR #394, opened 2022-06-01
Improve styling in new rule form

vincent-olivert-riera:
### Before

before

### After

after


vincent-olivert-riera:
I have added a few more commits to improve some other bits. The "after" screenshot has been updated accordingly.

vincent-olivert-riera:
I hope these would be the last additions.

Since the `rules_form_block.html` was reused in the "Edit Rule" page, the change of the "Basic" header to "New Rule" unconditionally was not OK. Now, thanks to 7522542 it displays the name of the rule when editing it, or "New Rule" when creating a new one.
Also in the "Edit Rule" page there was some strange rendering caused by the `input-group` class being applied to the `form` element. 06e502c drops that class and fixes the issue.

screenshot

vincent-olivert-riera:
> Looks good! Lets do one squash and then I can merge it

Rebased on `master` with fixup commits squashed.

PR #396, opened 2022-06-13
[BUGFIX] Fix handling of kwargs in skip_raw decorator

kfdm:
When calling post_save with Service, we call out to save_project, but
did not handle some of our required parameters correctly

- Change functions to all kwargs to hopefully catch errors in the future
- Default raw=False in our wrapper to match post_save.send


vincent-olivert-riera (approved):
LGTM

PR #403, opened 2022-09-22
Show rule description in rule page

vincent-olivert-riera:
Before, when you wanted to see the description of a rule, you needed to edit it in order to load the form and then look at the description field. That was a little bit weird and could be considered as an anti pattern for good usability.

Now, since we have added the description to the `rule_dict` template tag, it is displayed in the rule detail page.

### DIFF
--- a/promgen/templates/promgen/rule_detail.html
+++ b/promgen/templates/promgen/rule_detail.html
@@ -28,6 +28,12 @@ Promgen / Rule / {{ rule.name }}
 <div class="panel panel-primary">
     <div class="panel-heading">{{rule.name}}</div>
     <div class="panel-body">
+{% if rule.description %}
+        <div>
+            <label>Description:</label>
+            <p>{{rule.description}}</p>
+        </div>
+{% endif %}
         <pre v-pre>{{rule|rule_dict|pretty_yaml}}</pre>
     </div>
     <div class="panel-footer">


promgen/templatetags/promgen.py
```diff
@@ -36,6 +36,7 @@ def rule_dict(rule):
         "for": rule.duration,
         "labels": rule.labels,
         "annotations": rule.annotations,
+        "description": rule.description,
```
kfdm, line 39: This may be the wrong place for this, since this is trying to render in the Prometheus format. You should be able to edit the templates where it shows, and specifically show the description as something Promgen specific.

kfdm: It should probably go into the template around here

https://github.com/line/promgen/blob/d028f71346d6664a1cbc75b41db4caaf247ba56c/promgen/templates/promgen/rule_detail.html#L28-L37

kfdm (changes requested):

promgen/templatetags/promgen.py
vincent-olivert-riera: Fixed in a34503a.

promgen/templates/promgen/rule_detail.html
```diff
@@ -28,6 +28,10 @@ <h1>Rule: {{ rule.name }}</h1>
 <div class="panel panel-primary">
     <div class="panel-heading">{{rule.name}}</div>
     <div class="panel-body">
+        <div>
+            <label>Description:</label>
+            <p>{{rule.description}}</p>
+        </div>
         <pre v-pre>{{rule|rule_dict|pretty_yaml}}</pre>
     </div>
```
kfdm, line 36: Description is currently not a required field, so perhaps we should do this as a conditional .

```suggestion
{% if rule.description %}
    <div class="panel-body">
        <div>
            <label>Description:</label>
            <p>{{rule.description}}</p>
        </div>
        <pre v-pre>{{rule|rule_dict|pretty_yaml}}</pre>
    </div>
{% endif %}
```

So that we don't have just a label if there is no custom description.

kfdm (approved):
Looks good. Rebase and I can merge

vincent-olivert-riera:
Rebased on `master` with fixup commits squashed:
- I have amended the commit message accordingly
- I have fixed the position of the `if` to only wrap the description block and not the whole panel body.

PR #404, opened 2022-10-26
Exporter labels

Linux-yangyang:
git


kfdm:
This seems to be mostly a copy of a previously submitted PR but without any explanation. I'm going to close it in case this was submitted by mistake, but if you can make a case for the addition it can be re-opened.

PR #418, opened 2022-12-15
[INTERNAL] Migrate more tests to use fixtures

kfdm:
Previously we had several tests that used the Django ORM to populate data. Using fixtures should give us more reproducible test.

### DIFF
--- a/promgen/tests/test_cli.py
+++ b/promgen/tests/test_cli.py
@@ -11,46 +11,41 @@ from promgen.tests import PromgenTest
 
 
 class CLITests(PromgenTest):
+    fixtures = ["testcases.yaml", "extras.yaml"]
+
     @mock.patch("promgen.signals._trigger_write_config")
     def test_register_job(self, mock_signal):
         # Assert when project doesn't exist
         with self.assertRaises(CommandError):
             management.call_command("register-job", "missing-project", "example", 1234)
 
-        # Create a Service and Project and then try adding our job
-        shard = models.Shard.objects.create(name="TestShard")
-        service = models.Service.objects.create(name="TestService")
-        _ = models.Project.objects.create(name="TestProject", service=service, shard=shard)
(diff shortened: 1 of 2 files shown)


promgen/tests/test_cli.py
```diff
@@ -11,46 +11,41 @@
...
         # Assert when project doesn't exist
         with self.assertRaises(CommandError):
             management.call_command("register-job", "missing-project", "example", 1234)
 
-        # Create a Service and Project and then try adding our job
-        shard = models.Shard.objects.create(name="TestShard")
-        service = models.Service.objects.create(name="TestService")
-        _ = models.Project.objects.create(name="TestProject", service=service, shard=shard)
-        management.call_command("register-job", "TestProject", "example", 1234)
+        management.call_command("register-job", "test-project", "example", 1234)
```
vincent-olivert-riera, line 22: > We'll use cli- as a prefix for our objects to ensure they don't otherwise conflict with out fixture values.

In the `test_register_job` function we are using `test-project` and `example`. Shouldn't we do the same as in the `test_register_host` function and use `cli-project` and `cli.example.com`?

vincent-olivert-riera (approved):
I made one question, but overall looks good to me, so... approved 👍🏻

promgen/tests/test_cli.py
kfdm: So in one we're registering to an existing project (from the fixture) and one we are creating new jobs from the test. For the ones created by the test I added the `cli-` prefix. Should I try to reword things to be a bit more clear?

vincent-olivert-riera: > So in one we're registering to an existing project (from the fixture) and one we are creating new jobs from the test. For the ones created by the test I added the `cli-` prefix.

I see 👍🏻 

> Should I try to reword things to be a bit more clear?

No, I think it's fine as it is.

PR #419, opened 2022-12-20
Allow testing notifiers before registering them

vincent-olivert-riera:
This commit adds a new "Test" button next to the one for registering a notifier, so the user can use it to test if the notifier they are about to register actually works.

Example of success:

Screenshot 2022-12-20 at 19-04-30 Promgen 0 57 0 dev

Example of error:

Screenshot 2022-12-20 at 19-03-59 Promgen 0 57 0 dev

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -38,6 +38,16 @@ const silenceStore = {
     }
 };
 
+const notifierPreTestResultStore = {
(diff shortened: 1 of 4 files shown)


sh-kawakami:
@vincent-olivert-riera I think that this function we should implement and It looks fine. But please check these code can merge currently(because from make this PR, long time pasted.)

PR #424, opened 2023-06-09
[IMPROVEMENT] Rule Test should check for unrelated labels

kfdm:
A common source of confusion when writing Rules, is the global namespace of Prometheus when on a shared server. It's very easy to write a rule that covers more projects than expected.

To help avoid this, we want to upgrade the rule test button to check the labels that are returned vs the expected labels.

This was code that was written months ago, but got delayed due to other infra tasks. It could probably be made more robust, but I think we're ultimately going to take a deeper look at the rules system in the near future.


vincent-olivert-riera (approved):
Looks good to me 👍🏻

PR #425, opened 2023-07-03
Use default set of expected_labels

kfdm:
For Service and Project we have a specific set to test about, but when testing site rules (or potentially other types) there is not an expected set of labels we can test against. We need to still declare default expected_labels sets to avoid a Python error.

I ended up doing a little bit of reformatting as well in this PR, but if it's excessive I can redo this as two commits.


vincent-olivert-riera (approved):
Looks good to me 👍🏻

PR #426, opened 2023-07-05
Add debug warning to top of page

kfdm:
To make it easier to distinguish testing environment from a live one, we want to add an obvious warning alert at the top of the screen.

スクリーンショット 2023-07-05 14 45 03


vincent-olivert-riera (approved):
Looks good to me.

PR #427, opened 2023-07-13
Convert Rule label/annotation storage to JsonField

kfdm:
When Promgen was written, there was not a default JsonField, so labels and annotations were implemented as a class. Now that JsonField has been in Django since 3.1, it's a good time to change the implementation.

Previously, our audit log did *not* store change to labels/annotations (because they were separate objects), so one benefit of this migration is that those changes now show up in the edit log.

We also expect some queries to be a bit faster now that we do not need to query the other table relations for labels/annotations.

There may be some bits that could use better comments so feel free to point out the lines where the current explanation is insufficient.


vincent-olivert-riera (approved):
That was a lot of code and a long review. I didn't see anything suspicious of being broken, but sorry if I have missed it. To me, it looks good 👍🏻

PR #429, opened 2023-07-25
[INTERNAL] Upgrade to Python3.9

kfdm:
We will upgrade more dependencies later, but this PR attempts to upgrade just the Python version, and the minimum number of dependencies to get our builds working again.

We remove several pinned versions from requirements.in to better show some of our minimal requirements for the docker container. Our next PR will do some more updates of the requirements.


vincent-olivert-riera (approved):
LGTM

PR #432, opened 2023-07-31
[IMPROVEMENT] Additional alert filters

kfdm:
To help with debugging, it's useful to have a filter for both errors and other alerts that did not send. We also want to better highlight a date change to help with debugging.

### DIFF
--- a/promgen/templates/promgen/alert_list.html
+++ b/promgen/templates/promgen/alert_list.html
@@ -20,8 +20,8 @@
         <th>Service</th>
         <th>Project</th>
         <th>Job</th>
-        <th>Sent</th>
-        <th>Error</th>
+        <th><a href="?noSent=1">Sent</a></th>
+        <th><a href="?sentError=1">Error</a></th>
     </tr>
     {% for alert in alert_list %}
     <tr>
--- a/promgen/templates/promgen/alert_list.html
+++ b/promgen/templates/promgen/alert_list.html
@@ -24,6 +24,11 @@
         <th><a href="?sentError=1">Error</a></th>
     </tr>
     {% for alert in alert_list %}
+    {% ifchanged alert.created|date %}
+    <tr class="table-secondary">
+        <th colspan="8">{{ alert.created|date }}</th>
+    </tr>
+    {% endifchanged %}
     <tr>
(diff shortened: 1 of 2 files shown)


promgen/templates/promgen/alert_list.html
```diff
@@ -20,8 +20,8 @@
         <th>Service</th>
         <th>Project</th>
         <th>Job</th>
-        <th>Sent</th>
-        <th>Error</th>
+        <th><a href="?sent=0">Sent</a></th>
+        <th><a href="?error=1">Error</a></th>
```
vincent-olivert-riera, line 24: Why `sent` is set to `0` and `error` is set to `1`?

kfdm: Partly I was thinking of `zero sent emails` or `at least one error` but it probably does not matter significantly since we're only doing `filter()` or `exclude()` on the backend and not actually looking at the value.

kfdm: Do you have a better recommendation for a value to use for these ? I could also change the query parameter name.

vincent-olivert-riera: In both cases we are enabling a filter to only display alerts sent or alerts with errors, so I guess having `1` in both of them (`1` meaning `true`) looks better.

However, since we are not looking at the value, if the user manually changes the value and sets a `0` to disable the filter, it will not work as expected. So, I guess it would be better to look at the value.

Also, if you click on those links, the page reloads with a different url containing the new key, with the filter enabled. But, if I'm not mistaken, those links will appear again in the same way. What happens if you click again on the same one that you clicked before? I would expect the filter to be disabled. Is that the current behavior?

kfdm: Right now, none of the other search filters (service,project,etc) have a toggle either, so for now I'm just copying the behavior.

kfdm: I think the naming is too picky, but I've changed it in 6d9030e
`noSent` and `sentError` so they're more like flags.

vincent-olivert-riera: I guess it's fine by now.

vincent-olivert-riera (approved):

PR #433, opened 2023-08-04
[BUGFIX] Stricter checks about silencing global rules

kfdm:
Until we have a more robust permission system, we need to be cautious and ensure that users can not accidentally set a global level silence. To do this, we will check to see if the pending silence has alertname as a matcher, and if so, enforce that it also includes a service or project label.

### DIFF
--- a/promgen/tests/test_silence.py
+++ b/promgen/tests/test_silence.py
@@ -7,7 +7,7 @@ from unittest import mock
 from django.test import override_settings
 from django.urls import reverse
 
-from promgen import tests
+from promgen import forms, tests
 
 TEST_SETTINGS = tests.Data("examples", "promgen.yml").yaml()
 TEST_DURATION = tests.Data("examples", "silence.duration.json").json()
@@ -18,6 +18,8 @@ TEST_SETTINGS["timezone"] = "Asia/Tokyo"
 
 
 class SilenceTest(tests.PromgenTest):
+    fixtures = ["testcases.yaml", "extras.yaml"]
+
     def setUp(self):
         self.user = self.force_login(username="demo")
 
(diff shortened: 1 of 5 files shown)


promgen/tests/test_silence.py
```diff
@@ -61,3 +63,44 @@ def test_range(self, mock_post):
             )
 
         self.assertMockCalls(mock_post, "http://alertmanager:9093/api/v1/silences", json=TEST_RANGE)
+
+    @override_settings(PROMGEN=TEST_SETTINGS)
+    def test_site_silence_errors(self):
+        form = forms.SilenceForm(data={"labels": {}, "duration": "1m"})
+        self.assertEqual(
+            form.errors,
+            {"__all__": ["Unable to silence without labels"]},
```
vincent-olivert-riera, line 72: Uhm..., I have always found that asserting on exact sentences is very fragile.
For instance, that sentence is missing a period at the end. If I go and fix it in the source (where the exception is raised), then this test will break just because I fixed that sentence. Or if we decide to reword any of the error messages, their corresponding self tests will also break.
Moreover, if we ever add localization and decide to translate those error messages into multiple languages, it would make these assertions even more complicated.

I think it would be better to include an error code along with the message when the exception is raised. For instance:

```
raise forms.ValidationError(_("Unable to silence without labels"), code="ENOLABELS")
```

Then in the self tests retrieve the `ValidationError` object and assert on the `code`. For instance:

```
error_code = form.errors["__all__"].as_data()[0].code
self.assertEqual(error_code, "ENOLABELS")
```

vincent-olivert-riera (approved):
Overall looks good. I just left one suggestion for how the self tests do the assertions, but I will leave that decision entirely up to you, so I'm approving it :+1:

promgen/tests/test_silence.py
kfdm: I didn't realize that ValidationError supported a `code` parameter 🤔 
I agree, that would be a more reliable way to write a test 👍

kfdm: Seems like we might be able to combine Enums with Django's Validation error so we can more easily use error codes and also better centralize the localization strings for future use.

vincent-olivert-riera (approved):

PR #434, opened 2023-08-04
[INTERNAL] Use ruff for formatting

kfdm:
To reduce some of the extra tools and configuration, we want to
migrate to ruff as a replacement of isort and pyflakes.


vincent-olivert-riera (approved):
LGTM

PR #439, opened 2023-09-25
[INTERNAL] Remove envdir

kfdm:
Since we have django-environ and we have been moving to a docker deployment, it doesn't make sense to keep envdir around anymore. This PR cleans up the initial bits, though we'll also have a future PR to cleanup the `promgen bootstrap` command more.

### DIFF
--- a/promgen/__init__.py
+++ b/promgen/__init__.py
@@ -16,27 +16,3 @@ PROMGEN_CONFIG_DIR = pathlib.Path(
 PROMGEN_CONFIG_FILE = pathlib.Path(
     os.environ.get("PROMGEN_CONFIG", str(PROMGEN_CONFIG_DIR / "promgen.yml"))
 )
-
-# In the future we want to remove the dependency on envdir, but this will
-# require some cleanup and documentation changes, so for now we will just
-# ensure that it can't fail here if envdir is not already installed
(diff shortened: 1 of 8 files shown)


promgen/__init__.py
```diff
@@ -16,27 +16,3 @@
...
-    try:
-        from django.utils.crypto import get_random_string
-        import warnings
-
-        warnings.warn("Unset SECRET_KEY setting to random for now")
-        os.environ["SECRET_KEY"] = get_random_string(
-            50, "abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)"
-        )
-    except ImportError:
-        pass
```
vincent-olivert-riera, line 42: After removing this the CI is failing with:

```
1.378 django.core.exceptions.ImproperlyConfigured: Set the SECRET_KEY environment variable
```

Perhaps we should set a random `SECRET_KEY` env variable in the GitHub Actions workflow file.

vincent-olivert-riera (changes requested):

promgen/__init__.py
kfdm: I'm setting it in the docker build step for now.

vincent-olivert-riera (approved):

PR #441, opened 2023-10-30
Convert Vue filters into methods

vincent-olivert-riera:
In Vue 3, filters are removed and no longer supported. Instead, they recommend replacing them with method calls or computed properties.

https://v3-migration.vuejs.org/breaking-changes/filters.html

Since those methods can be reused in multiple components we store them as mixins instead of defining them in specific components. Mixins work in both Vue 2 and Vue 3.

The recommendation for migrating global filters in Vue 3 is to store them as global properties [1], however that is not supported in Vue 2.

1: https://vuejs.org/api/application.html#app-config-globalproperties

---

This PR will make the future PR for migrating to Vue 3 much shorter.

### DIFF
--- a/promgen/static/js/mixins.vue.js
+++ b/promgen/static/js/mixins.vue.js
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2023 LINE Corporation
+ * These sources are released under the terms of the MIT license: see LICENSE
+ */
+
+var mixins = {
+  methods: {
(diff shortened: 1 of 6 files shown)


kfdm (dismissed):
LGTM

vincent-olivert-riera:
I fixed the year in the copyright of the `mixins.vue.js` file.

kfdm (approved):

PR #442, opened 2023-10-30
Migrate to Vue 3

vincent-olivert-riera:
I have explained all the changes and their reasons in the commit message.
I couldn't split this into different commits without breaking anything.


promgen/static/js/promgen.vue.js
```diff
@@ -3,26 +3,24 @@
  * These sources are released under the terms of the MIT license: see LICENSE
  */
 
-Vue.config.devtools = true
```
kfdm, line 6: How does one set devtools mode to debug with the VueJS extension now ?

vincent-olivert-riera: > We started importing the production build of Vue [...] When doing development we (can) easily change that manually.

There is a missing "can" in that sentence, but basically those are the instructions: manually changing the import whenever you need to do UI development.

Instead of using the production version, you use the normal one. In Vue 3 the dev tools are enabled by default, there is no need to explicitly enabling them with a code instruction. Perhaps the misunderstanding came from there? Should I elaborate more that part of the commit message?

kfdm: I think for completeness sake, there needs to be an explanation of where to do this. But at a glance this would involve manually downloading the correct vuejs build and putting it in the right place and updating the import.

Feels like we should probably drop the dev version of vue.js in there, and perhaps toggle based on `settings.DEBUG` or `{% if debug == True %}` in the template (I didn't look up the exact syntax for this comment)

vincent-olivert-riera: Fine by me. I'll do the changes 👍🏻

vincent-olivert-riera:
Changes done. Needs rebasing to:
- Reorder commits.
- Squash fixups.
- Amend commit message.

kfdm (approved):
This all looks good. Lets do a rebase for final approval :)

vincent-olivert-riera:
> This all looks good. Lets do a rebase for final approval :)

Done 👍🏻

PR #444, opened 2023-10-31
Fix github pages deploy

### LINKED ISSUE #443: Documentation is not updated since 2021
Hello, I'm trying to use promgen and looked around documentation a bit, and noticed something is off.

# What's wrong?

- Current gh-pages release: https://line.github.io/promgen/

Looks like current documentation is published in Mar 30, 2021 and it is not updated since then.

image

And in [sphinx.yml](https://github.com/line/promgen/blob/8ad2ed7c3998c8e1212e4b1bb17998662ed14aef/.github/workflows/sphinx.yml#L36), it checks github reference if it is same with `refs/heads/main`, which should be actually `refs/heads/master`.

So [`Deploy to GitHub Pages`](https://github.com/line/promgen/blob/8ad2ed7c3998c8e1212e4b1bb17998662ed14aef/.github/workflows/sphinx.yml#L35) step is always skipped, so that documentation is not published.

# How to fix

image

Checking if github ref is `master` is an easy fix, but it's not enough.

`gh-pages` branch is not necessary, because in github, it is possible to directly deploy 'pages' with github actions (⚠️ if only changed repository settings)

In this case, generated files are attached as "Artifacts" in actions, and it is possible to deploy Artifacts with [`actions/deploy-pages`](https://github.com/actions/deploy-pages), so no need to use third-party action (e.g. `JamesIves/github-pages-deploy-action`)

I will make PR shortly, but if you have another opinion, please let me know!

sh-cho:
Closes #443 

- deploy with action artifacts, not from gh-pages branch
- check ref with master, not with main

ex) https://github.com/sh-cho/promgen/actions/runs/6706997395

Only deploy on push master branch (using [github actions context](https://docs.github.com/en/actions/learn-github-actions/contexts#determining-when-to-use-contexts))

`gh-pages` branch can be removed after this PR is merged

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -44,7 +44,7 @@ jobs:
 
   deploy:
     needs: build
-    if: ${{ github.ref == 'refs/heads/master' }}
+    if: ${{ github.ref_name == github.event.repository.default_branch }}
 
     permissions:
       pages: write
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -7,8 +7,13 @@ on:
     branches:
       - master
 
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
 jobs:
-  docs:
+  build:
     name: Build Sphinx Documentation
     runs-on: ubuntu-20.04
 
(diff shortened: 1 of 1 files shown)


.github/workflows/sphinx.yml
```diff
@@ -32,10 +37,25 @@ jobs:
...
+        uses: actions/upload-pages-artifact@v2
         with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          BRANCH: gh-pages
-          FOLDER: dist/html
+          path: 'dist/html'
+
+  deploy:
+    needs: build
+    if: ${{ github.ref == 'refs/heads/master' }}
```
kfdm, line 47: I wonder if it would be better to use the variable for this 🤔 
```suggestion
    if: ${{ github.ref == github.event.repository.default_branch }}
```

sh-cho: I think we need to compare with `refs/heads/{default_branch}`

(https://stackoverflow.com/a/69995861/4295499)

kfdm: Ah, I guess there is also `github.ref_name` that could work in this case.

```suggestion
    if: ${{ github.ref_name == github.event.repository.default_branch }}
```

Could work. Your solution also works, but I think I sometimes over think trying to find the simplest solution 😆

sh-cho: @kfdm gentle ping :)

kfdm: less than 24 hours is a bit soon for a ping. Sometimes other work tasks get in the way of checking pull requests 😅

sh-cho: haha sorry 😅

kfdm (approved):
Looks good to me! 

Thank you very much for catching this. I think we (as a team) never quite decided on github pages or readthedocs for the documentation and unfortunately it looks like both methods broke over time 😅 

With this PR, I think we have a much better setup and we can probably use the github pages version for our docs. Thank you for your contribution 🙇

PR #446, opened 2023-11-02
Configure concurrency setting for actions

kfdm:
We can set a concurrency setting for each of our actions so that pushing to a branch cancels in-progress runs without conflicting with runs on other branches

https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -8,6 +8,10 @@ on:
     branches:
       - master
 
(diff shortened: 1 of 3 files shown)


vincent-olivert-riera (approved):
LGTM 👍🏻

PR #455, opened 2023-11-09
Dependabot

kfdm:
Collecting dependabot updates into a single PR.

Looking at some tags as an example
https://github.com/actions/setup-python/tags
We have to pin more exact versions since there's not really a concept of semver for tags


vincent-olivert-riera (approved):
LGTM 👍🏻

PR #456, opened 2023-11-15
Add build pipeline for Primevue

bai1024:
This PR shows the build pipeline for [Primevue](https://primevue.org/)
Before try to use the new components, we need to import them first in `promgen-gen/main.js`
Then run `make primevue` command.
After the new file is generated under `static/promgen`. You can start to use the new components!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@
 build
 coverage.xml
 dist
+primevue-gen/node_modules
+package-lock.json
+promgen/static/primevue/
 
 # Prometheus/Docker ignores
 docker/promgen.json
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,8 +19,8 @@ The back end is written in Python. The front end has some parts that require a b
 ```bash
 # If you need to install Python or Node.js, try using your system's package manager
 # Examples
-# yum install python3 python3-pip nodejs
-# homebrew install python3
+# yum install python3 python3-pip nodejs # On RedHat
+# homebrew install python3 node # On MacOS
 # If using OSX with Homebrew, you may need to export some flags
 # to get mysql client to install
 # export LDFLAGS="-I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ You can see some of the commands by running `make help`
 If you are using an external database, you will want to create a `promgen` and `test_promgen`
 database before running bootstrap or `make migrate` and `make test` will fail.
 
-The back end is written in Python. The front end has some parts that require a build step, and for that Node.js is required.
+Promgen's backend is written in Python with parts of it's frontend written in Vuejs. Both Python and Node.js are required for development.
 
 ```bash
 # If you need to install Python or Node.js, try using your system's package manager
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ database before running bootstrap or `make migrate` and `make test` will fail.
 Promgen's backend is written in Python with parts of it's frontend written in Vuejs. Both Python and Node.js are required for development.
 
 ```bash
-# If you need to install Python or Node.js, try using your system's package manager
+# Python and Node.js can be installed with your system's package manager.
 # Examples
 # yum install python3 python3-pip nodejs # On RedHat
 # homebrew install python3 node # On MacOS
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,10 +14,12 @@ You can see some of the commands by running `make help`
 If you are using an external database, you will want to create a `promgen` and `test_promgen`
 database before running bootstrap or `make migrate` and `make test` will fail.
 
+The back end is written in Python. The front end has some parts that require a build step, and for that Node.js is required.
+
 ```bash
-# If you need to install Python first, try using your system's package manager
+# If you need to install Python or Node.js, try using your system's package manager
 # Examples
-# yum install python3 python3-pip
+# yum install python3 python3-pip nodejs
 # homebrew install python3
 # If using OSX with Homebrew, you may need to export some flags
 # to get mysql client to install
@@ -32,6 +34,8 @@ echo 1 > ~/.config/promgen/DEBUG
 make migrate
 # Run tests
 make test
+# Build front end
+make primevue
 # Run development server
 make run
 ```
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ migrate: $(APP_BIN)
 
 .PHONY:	run
 ## Django: Run development server
-run: migrate
+run: migrate primevue
 	$(APP_BIN) runserver
 
 .PHONY: shell
@@ -163,3 +163,12 @@ clean:
 .PHONY: changelog
 changelog:
 	git log --color=always --first-parent --pretty='format:%s|%Cgreen%d%Creset' | column -ts '|' | less "$(lessopts)"
+
+.PHONY: primevue
+primevue: primevue-gen/node_modules promgen/static/primevue/main.css promgen/static/primevue/main.js
+
+primevue-gen/node_modules: primevue-gen/package.json
+	cd primevue-gen && npm i --no-package-lock
+
+promgen/static/primevue/main.css promgen/static/primevue/main.js: primevue-gen/build.js primevue-gen/src/main.js
+	cd primevue-gen && npm run build
--- a/primevue-gen/build.js
+++ b/primevue-gen/build.js
@@ -1,0 +1,30 @@
+// https://esbuild.github.io/plugins/#using-plugins
+
+const esbuild = require("esbuild");
+
+const myPlugin = {
+  name: "my-plugin",
+  setup(build) {
+    build.onResolve({ filter: /^vue$/ }, (args) => ({
+      path: args.path,
+      namespace: "my-plugin",
+    }));
+
+    build.onLoad({ filter: /.*/, namespace: "my-plugin" }, () => ({
+      contents: "module.exports = Vue",
+    }));
+  },
+};
+(async () => {
+  await esbuild.build({
+    entryPoints: ["./src/main.js"],
+    bundle: true,
+    outdir: "../promgen/static/primevue",
+    logLevel: "info",
+    loader: {
+      ".woff2": "dataurl",
+    },
+    plugins: [myPlugin],
+    minify: true,
+  });
+})();
--- a/primevue-gen/package.json
+++ b/primevue-gen/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "primvue-gen",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "build": "node build.js"
+  },
+  "keywords": [],
+  "author": "",
+  "dependencies": {
+    "primevue": "3.40.1"
+  },
+  "devDependencies": {
+    "esbuild": "0.19.5"
+  }
+}
--- a/primevue-gen/src/main.js
+++ b/primevue-gen/src/main.js
@@ -1,0 +1,10 @@
+import Vue from "vue";
+import PrimeVue from "primevue/config";
+import "primevue/resources/themes/bootstrap4-light-blue/theme.css";
+
+window.createPrimeVueApp = function (opt) {
+  const app = Vue.createApp(opt);
+  app.use(PrimeVue);
+
+  return app;
+};
--- a/promgen/templates/base.html
+++ b/promgen/templates/base.html
@@ -11,6 +11,13 @@
   <link rel="stylesheet" href="{% static 'css/bootstrap-switch.min.css' %}">
   <link rel="stylesheet" href="{% static 'css/promgen.css' %}">
   <link rel="icon" href="{% static 'images/promgen_logo_color.png' %}">
+  {% if debug %}
+  <script src="{% static 'js/vue-3.3.7/vue.global.js' %}"></script>
+  {% else %}
+  <script src="{% static 'js/vue-3.3.7/vue.global.prod.js' %}"></script>
+  {% endif %}
+  <script src="{% static 'primevue/main.js' %}"></script>
+  <link rel="stylesheet" href="{% static 'primevue/main.css' %}" />
 </head>
 
 <body>
@@ -35,11 +42,6 @@
   <script src="{% static 'js/luxon.min.js' %}"></script>
   <script src="{% static 'js/linkify.min.js' %}"></script>
   <script src="{% static 'js/linkify-string.min.js' %}"></script>
-  {% if debug %}
-  <script src="{% static 'js/vue-3.3.7/vue.global.js' %}"></script>
-  {% else %}
-  <script src="{% static 'js/vue-3.3.7/vue.global.prod.js' %}"></script>
-  {% endif %}
   <script type="text/x-template" id="bootstrap-panel-template">
     {% include 'promgen/vue/bootstrap_panel.html' %}
   </script>
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -39,7 +39,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build and push
-        uses: docker/build-push-action@v5.0.0
+        uses: docker/build-push-action@v5.1.0
         with:
           context: .
           push: ${{ github.event_name != 'pull_request' }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,13 @@
+# Build frontend
+FROM node:20.10.0-alpine as frontend-builder
+WORKDIR /build
+COPY primevue-gen ./primevue-gen
+RUN mkdir -p promgen/static
+RUN set -ex \
+    && cd primevue-gen \
+    && npm install \
+    && npm run build
+
 FROM python:3.9-alpine
 LABEL maintainer=paul.traylor@linecorp.com
 
@@ -44,6 +54,7 @@ COPY docker/docker-entrypoint.sh /
 COPY setup.py /usr/src/app/setup.py
 COPY setup.cfg /usr/src/app/setup.cfg
 COPY promgen /usr/src/app/promgen
+COPY --from=frontend-builder /build/promgen/static/primevue /usr/src/app/promgen/static/primevue
 COPY promgen/tests/examples/promgen.yml /etc/promgen/promgen.yml
 
 WORKDIR /usr/src/app
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -46,7 +46,7 @@ const exporterTestResultStore = Vue.reactive({
     },
 });
 
-const app = Vue.createApp({
+const app = createPrimeVueApp({
     delimiters: ['[[', ']]'],
     data() {
         return dataStore;


promgen/templates/base.html
```diff
@@ -1,83 +1,88 @@
 {% load static %}
 <!DOCTYPE html>
 <html>
-
-<head>
```
bai1024, line 5: Updated this file's format, so this file has a lot of changes

Makefile
```diff
@@ -163,3 +163,7 @@ clean:
 .PHONY: changelog
 changelog:
 	git log --color=always --first-parent --pretty='format:%s|%Cgreen%d%Creset' | column -ts '|' | less "$(lessopts)"
+
+.PHONY: primevue
+primevue:
+	cd primevue-gen && npm i && npm run build
```
vincent-olivert-riera, line 169: This can be improved so the build happens only when necessary. Look at the code below:

```
.PHONY: primevue
# By building primevue the following objects will be generated:
# - primevue-gen/node_modules
# - promgen/static/primevue/main.css
# - promgen/static/primevue/main.js
# Therefore, they should be added as dependencies.
#
# Also, let's create rules for each one of those dependencies, so they will be built
# when they are missing. This primevue rule does nothing but listing dependencies.
primevue: primevue-gen/node_modules promgen/static/primevue/main.css promgen/static/primevue/main.js

# The node_modules object depends on the package.json, so let's add it to the
# dependencies. That will make this rule to be executed if the package.json changes.
# The only thing needed for building the node_modules object is "npm i".
primevue-gen/node_modules: primevue-gen/package.json
       cd primevue-gen && npm i

# The main.css and main.js objects depend on the build.js and src/main.js files, so
# let's add them to the dependencies. If any of them changes, this rules will be
# executed.
# The only thing needed to build those objects is "npm run build".
promgen/static/primevue/main.css: primevue-gen/build.js primevue-gen/src/main.js
promgen/static/primevue/main.js: primevue-gen/build.js primevue-gen/src/main.js
       cd primevue-gen && npm run build
```

Comments are just for illustration purposes, we don't need to add them.

primevue-gen/build.js
```diff
@@ -0,0 +1,29 @@
...
+    entryPoints: ["./src/main.js"],
+    bundle: true,
+    outdir: "../promgen/static/primevue",
+    logLevel: "info",
+    loader: {
+      ".woff2": "dataurl",
+    },
+    plugins: [myPlugin],
+  })
+})()
```
vincent-olivert-riera, line 29: A few general comments about this file (`primevue-gen/build.js`):
- Plenty of semi-colons missing; Please add them.
- Some semi-colons are misplaced (look at the begging of the `async` line).
- The file is missing a new-line character. Please add it.
- Has this file been written following some kind of documentation about how to write your build script for esbuild? If so, it would be nice to include a link at the top of this file in case we need to modify it in the future.

primevue-gen/build.js
```diff
@@ -0,0 +1,29 @@
...
+  setup(build) {
+    build.onResolve({ filter: /^vue$/ }, (args) => ({
+      path: args.path,
+      namespace: "my-plugin",
+    }))
+
+    build.onLoad({ filter: /.*/, namespace: "my-plugin" }, () => {
+      const contents = "module.exports = Vue"
+      return { contents }
+    })
```
vincent-olivert-riera, line 14: This is the same, but shorter and without needing to define a variable and a explicit return expression:
```
build.onLoad({ filter: /.*/, namespace: "my-plugin" }, () => ({
  contents: "module.exports = Vue"
}))
```
Moreover, it is also more consistent since it's exactly the same thing you did before in `build.onResolve`: taking advantage of the implicit return value of the arrow function.

primevue-gen/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "node build.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "esbuild": "0.19.5",
```
vincent-olivert-riera, line 13: `esbuild` is not a run-time dependency. It is a build-time dependency so we should place it in a `devDependencies` block:
```
  "dependencies": {
    "primevue": "3.40.1"
  },
  "devDependencies": {
    "esbuild": "0.19.5"
  }
```

primevue-gen/src/main.js
```diff
@@ -0,0 +1,22 @@
...
+  const app = Vue.createApp(opt)
+  app.use(PrimeVue)
+
+  app.component('p-dropdown', Dropdown);
+  app.component('p-button', Button);
+  app.component('p-menubar', Menubar);
+  app.component('p-avatar', Avatar);
+  app.component('p-badge', Badge);
+  return app
+}
```
vincent-olivert-riera, line 22: General comments about this file (`primevue-gen/src/main.js`):
- Plenty of semi-colons missing; please add them.
- The file is missing a new-line character; please add it.
- It would be nice to have all the Primevue imports and component registrations alphabetically sorted.

primevue-gen/src/main.js
```diff
@@ -0,0 +1,22 @@
+import Vue from "vue"
+import PrimeVue from "primevue/config"
+import 'primevue/resources/themes/lara-light-teal/theme.css'
```
vincent-olivert-riera, line 3: Shouldn't we use the Bootstrap theme instead?

primevue-gen/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "primvue-gen",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
```
vincent-olivert-riera, line 5: There isn't any `index.js` file, and this still works just fine. I wonder if we could just remove this line.

primevue-gen/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "scripts": {
+    "build": "node build.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
vincent-olivert-riera, line 11: Shouldn't we use the same license that we use for Promgen? (MIT)
In fact, is this field actually necessary? If it can be removed, I would prefer that, because we already have a `LICENSE` file at the root of the repo.

promgen/templates/base.html
```diff
@@ -1,83 +1,88 @@
...
+      <option>critical</option>
+      <option>major</option>
+      <option>minor</option>
+      <option>debug</option>
+    </datalist>
 
+    <script>
+      app.mount("#vue");
+    </script>
+  </body>
```
vincent-olivert-riera, line 87: > Updated this file's format, so this file has a lot of changes.
> The main changes of base.html is to move the import of vuejs to the top of this file.

If you want to place the Vue imports at the top of this file, just do that.
If you want to reformat the file, then do it in a separate commit.

Both things are unrelated and don't depend on each other, so they should not be in the same commit. That makes the `git diff` unnecessarily bigger, and harder to review when looking for bugs.

primevue-gen/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "dependencies": {
+    "esbuild": "0.19.5",
+    "primevue": "3.40.1"
+  }
+}
```
vincent-olivert-riera, line 16: General comments about this file (`primevue-gen/package.json`) and also `package-lock.json`:
- Since we are pinning the versions of the packages directly in this file, I think we should get rid of the `package-lock.json` file by running `npm i --no-package-lock`. Also we should add `package-lock.json` to `.gitignore`.
- I don't know if you noticed, but the content of the `package-lock.json` file has internal URLs on it, probably because you have set a local mirror for NPM repositories. Thouse URLs would not work for anyone outside your company (Promgen is an open source project). Since other people may also want to use their own NPM mirrors, I think it's another reason for not generating (and commit) a `package-lock.json` file.

.gitignore
```diff
@@ -14,6 +14,7 @@
 build
 coverage.xml
 dist
+primevue-gen/node_modules
```
vincent-olivert-riera, line 17: I think `promgen/static/primevue/` should also be added to `.gitignore`, since it will be generated by the build pipeline. It will be included in the released tarballs, but it should not be commit into the repo.

vincent-olivert-riera (changes requested):
Review done.

Looks very good in general. Thanks 👍🏻

primevue-gen/src/main.js
bai1024: > It would be nice to have all the Primevue imports and component registrations alphabetically sorted.

I think it's best to selectively import dependencies as needed during our UI update process. This approach can help reduce the size of the bundle

.gitignore
bai1024: 🤔  I think we don't need to add that to `.gitignore`, since we only use this command to generate file only when we want to import new components.

kfdm: Generally I would agree that generated files should not be added, but should we consider the case of a python developer who wanted to work on the django bits, but not touch the javascript bits. Should we require them to also have node installed to get a working environment to develop in 🤔 
I think I'm slightly in favor of committing the generated files to `promgen/static/primeview` similar to how we have the binary files from vue and bootstrap and such downloaded from a release archive 🤔

primevue-gen/src/main.js
vincent-olivert-riera: There has been a misunderstanding in here. I didn't want you to import all the stuff provided by Primevue.

I meant that we could convert this...
```
import Button from 'primevue/button';
import Dropdown from "primevue/dropdown";
import Menubar from 'primevue/menubar';
import Avatar from 'primevue/avatar';
import Badge from "primevue/badge";
```
...into this:
```
import Avatar from 'primevue/avatar';
import Badge from "primevue/badge";
import Button from 'primevue/button';
import Dropdown from "primevue/dropdown";
import Menubar from 'primevue/menubar';
```
So it's alphabetically sorted, looks more organized and is easier to find stuff within that list.

Same for the component registration, converting this...
```
  app.component('p-dropdown', Dropdown);
  app.component('p-button', Button);
  app.component('p-menubar', Menubar);
  app.component('p-avatar', Avatar);
  app.component('p-badge', Badge);
```
...into this:
```
  app.component('p-avatar', Avatar);
  app.component('p-badge', Badge);
  app.component('p-button', Button);
  app.component('p-dropdown', Dropdown);
  app.component('p-menubar', Menubar);
```

.gitignore
vincent-olivert-riera: > should we consider the case of a python developer who wanted to work on the django bits, but not touch the javascript bits. Should we require them to also have node installed to get a working environment to develop in 

I don't think requiring to install node on a dev machine is something necessarily bad. It is a dev machine after all.

> I think I'm slightly in favor of committing the generated files to promgen/static/primeview similar to how we have the binary files from vue and bootstrap and such downloaded from a release archive 🤔

I get your point, but I'm more in favor of not committing the generated files for the following reasons:
- The Vue and Bootstrap libraries that we include come from official releases of those projects. They are public and well known, and they are not subject to change unless we (internally) decide to update to new releases. Nobody will ever modify those libraries that we include in our repo.
- The generated `main.css` and `main.js` files do not come from a trusty source. Instead, they are the result of a command line execution. And they are subject to change by anyone. For instance, I could imagine someone from anywhere in the world doing a PR for changes in the UI, which involve a new import for a Primevue component that we do not use yet. That would modify the generated `main.js`, which is around 7000 lines of JavaScript. Reviewing that would not hard. In fact, if we end up minifying that file, which I think we should do for performance reasons, it would be not just hard but impossible.
- The generated `main.css` and `main.js` file do not just bundle code from the Primevue library. Right now they do, yes. But in the future they will also bundle compiled single file component (`.vue`) files. Reviewing such compiled JavaScript code is hard or impossible, and the PRs could come from anywhere (untrusted sources), so they should be reviewed. Who knows what kind of bad stuff could be injected in an unreviewable soup of generated JavaScript code.

If you really want to go with committing those files, fine, I would not object more. But if I put those reasons on the left plate of a scale, and installing node on the right plate, I would choose the right plate.

promgen/templates/base.html
vincent-olivert-riera: It looks like you have addressed this comment in the following commits:
- 43381cccbaa8ee644d3c72c043c22c552fec6586
- ee0b9c925ac14b7a152c00e9b6c48f3a7bb1e0e7
- a3677058a4338f0497c9bccdfd7176bbdc13b3c5

This conversation is resolved.

primevue-gen/package.json
vincent-olivert-riera: This is not completely resolved:
- You ignored `package-lock.json` with 492ccde551746518465a0c7d0b929098ca871350.
- You removed `package-lock.json` with bfb968b0b33a3f0803f338c0c23ba2557147c604.

But the `npm` command stills needs to be addressed to avoid generating that file: `npm i --no-package-lock`.

primevue-gen/package.json
vincent-olivert-riera: I see you have addressed this commit with 43381cccbaa8ee644d3c72c043c22c552fec6586.

This conversation is now resolved.

primevue-gen/package.json
vincent-olivert-riera: I see you have addressed this commit with 43381cccbaa8ee644d3c72c043c22c552fec6586.

This conversation is now resolved.

primevue-gen/src/main.js
vincent-olivert-riera: @bai1024 , was that comment missed? I didn't see any code change to address it, or a reply to it.

primevue-gen/package.json
vincent-olivert-riera: I see you have addressed this commit with 43381cccbaa8ee644d3c72c043c22c552fec6586.

This conversation is now resolved.

primevue-gen/build.js
vincent-olivert-riera: I see you have addressed this commit with 43381cccbaa8ee644d3c72c043c22c552fec6586.

This conversation is now resolved.

primevue-gen/build.js
vincent-olivert-riera: > - Plenty of semi-colons missing; Please add them.
> - Some semi-colons are misplaced (look at the begging of the async line).
> - The file is missing a new-line character. Please add it.

I see you have addressed this comment with 43381cccbaa8ee644d3c72c043c22c552fec6586.

> - Has this file been written following some kind of documentation about how to write your build script for esbuild? If so, it would be nice to include a link at the top of this file in case we need to modify it in the future.

I see you have addressed this comment with 3a760cdf029d75a44d089bf67dda4f2a4314f48e.

This conversation is now resolved.

vincent-olivert-riera (changes requested):
@bai1024 , I have reviewed your changes and resolved the conversations that were fully addressed.

Some of them are still unresolved, so please have a look.

For the next review spin, feel free to squash all your current commits before adding new ones.

primevue-gen/build.js
```diff
@@ -0,0 +1,29 @@
...
+
+    build.onLoad({ filter: /.*/, namespace: "my-plugin" }, () => ({
+      contents: "module.exports = Vue",
+    }));
+  },
+};
+(async () => {
+  await esbuild.build({
+    entryPoints: ["./src/main.js"],
+    bundle: true,
```
vincent-olivert-riera, line 21: I think we should also add `minify: true,`. It will save some bandwidth and help Promgen load faster.

Non-minified:
```
  ../promgen/static/primevue/main.css  789.7kb
  ../promgen/static/primevue/main.js   311.4kb
```

Minified:
```
  ../promgen/static/primevue/main.css  756.7kb
  ../promgen/static/primevue/main.js   170.6kb
```

At the end of the day, the generated non-minified JavaScript code is already an unreadable mess, so the smaller it is, the better 🤷🏻 .

primevue-gen/src/main.js
bai1024: Sorry, this thread was hidden.
The components in design is based on `lara` theme instead of `Bootstrap` theme.

primevue-gen/build.js
bai1024: addressed this comment in [8f92674](https://github.com/line/promgen/pull/456/commits/8f926746528ac8ad48958a64a23ba8f0bce8e762)

primevue-gen/src/main.js
bai1024: Addressed this comments in [48351b1](https://github.com/line/promgen/pull/456/commits/48351b11fe2612ccdc7f61b133241482bb75d52e)

primevue-gen/package.json
bai1024: Addressed the comment in [f4c9367](https://github.com/line/promgen/pull/456/commits/f4c9367e5793a87e9da76026f37268a149bceecd)

Makefile
bai1024: addressed this comments in [b821c93](https://github.com/line/promgen/pull/456/commits/b821c93d281f6f6ad4cd9fa4c3a0e30f2a793043)

primevue-gen/src/main.js
vincent-olivert-riera: Uhm..., maybe it was a misunderstanding from my side, but I thought the idea was to use the Bootstrap theme from Primevue because Promgen currently uses Bootstrap, so that would allow us to progressively change elements from the UI maintaining the current style, which would feel more natural.

Maybe we should discuss this withing the team.

.gitignore
sh-kawakami: I agree that builded codes from others are untrusted.
If we have to be considerate of the developer who don't want to set up node environment, we have a mean that we build `main.css` and `main.js` by us in advance I think.

primevue-gen/src/main.js
bai1024: addressed this comments in [633b001](https://github.com/line/promgen/pull/456/commits/633b001ac6672e13beb2718892d50c126073216d)

vincent-olivert-riera (commented):
@bai1024, you have done a fantastic job addressing all my comments 👍🏻 Perhaps it would be a good moment to squash all your commits for a final check.

vincent-olivert-riera (approved):
I've done the last check after the squash, and looks good to me.
I don't have any more comments, so from my side this PR is approved.

The [question](https://github.com/line/promgen/pull/456/files#r1395061533) about the generated file still remains. I already gave my opinion, but ultimately the decision is not mine.

.gitignore
kfdm: The python developer in me doesn't like the idea of requiring node just to run Promgen for development, but to avoid issues with old binary versions and untrusted binaries in pull requests, it's likely best that we don't check in the build outputs and require node for development environments as well.

- We need to ensure that the Makefile properly checks requirements (`make run` should probably ensure the javascript files exist)
- We need to ensure that the docker build creates the required javascript files

vincent-olivert-riera:
> I'm still needing to find time to review this, but remembered we will also need to ensure this still works with a docker container. In that case, we likely want a muti build image that builds the javascript in it's own image and then copies it over to the release image.

@bai1024 , I have created a [pull request](https://github.com/bai1024/promgen/pull/1) for the necessary changes to the Makefile and the Dockerfile. After your merge my commits into your branch, you can squash them all together and update this PR.

Also, I noticed you tried to remove the generated files from this PR with this commit: f70e159. However, I think during the rebase you did after that the commit was removed and the generated files are back on this PR. Please check it out.

.gitignore
bai1024: add this comments at [e699653](https://github.com/line/promgen/pull/456/commits/e699653ff33c1b70692fb25dc1a9a340aa13bc6f)

CONTRIBUTING.md
```diff
@@ -14,10 +14,12 @@ You can see some of the commands by running `make help`
...
 
+The back end is written in Python. The front end has some parts that require a build step, and for that Node.js is required.
+
 ```bash
-# If you need to install Python first, try using your system's package manager
+# If you need to install Python or Node.js, try using your system's package manager
 # Examples
-# yum install python3 python3-pip
+# yum install python3 python3-pip nodejs
 # homebrew install python3
```
kfdm, line 23:
```suggestion
# yum install python3 python3-pip nodejs # On RedHat
# homebrew install python3 node # On MacOS
```

CONTRIBUTING.md
```diff
@@ -14,10 +14,12 @@ You can see some of the commands by running `make help`
 If you are using an external database, you will want to create a `promgen` and `test_promgen`
 database before running bootstrap or `make migrate` and `make test` will fail.
 
+The back end is written in Python. The front end has some parts that require a build step, and for that Node.js is required.
+
```
kfdm, line 18:
```suggestion
Promgen's backend is written in Python with parts of it's frontend written in Vuejs. Both Python and Node.js are required for development.

```

kfdm (commented):
Some minor suggestions on rewording the contributing.md additions.

kfdm:
Internally we discussed handling some of this a different way, so will close this PR for now.

PR #458, opened 2023-11-28
Fix "make run" without debug mode

vincent-olivert-riera:


kfdm (commented):
[Partly a quesiton](https://github.com/line/promgen/blob/a2b6412d985870c96cb3b7f857082e6584cab46e/promgen/settings.py#L41)

Just a confirmation, but if we assume a django only developer that wrote `DEBUG=1` into their `./.env` file then this would not have any effect, only if a user runs `export DEBUG=1` (or similar) into their shell. Is this something we're worried about confusing the user with ? I'm not aware of there being a simple way to have `make` pick up a `.env` file for this edge case.

https://github.com/line/promgen/blob/a2b6412d985870c96cb3b7f857082e6584cab46e/promgen/settings.py#L167C6-L167C6
Alternatively is there some setting we could be using here that would avoid the need for collectstatic when running with dev ?

Makefile
```diff
@@ -114,6 +114,9 @@ migrate: $(APP_BIN)
 .PHONY:	run
 ## Django: Run development server
 run: migrate
+ifneq ($(DEBUG),1)
```
kfdm, line 117: I think I misread `ifneq` as `if` 😅

kfdm (approved):

PR #459, opened 2023-11-29
Fix Promgen init documentation

frakev:
Value of case is **docker-compose-boostrap** not **boostrap**
https://github.com/line/promgen/blob/master/docker/docker-entrypoint.sh#L10


kfdm (approved):
Thanks for the documentation fix! 🙇

PR #472, opened 2024-01-19
Audit logs: Show projects for services

vincent-olivert-riera:
Before, when displaying the audit logs for services, there was no information regarding when a project was added on a service, or modified, or deleted.

That information is useful when tracking things down while debugging, so let's add it.

### DIFF
--- a/promgen/views.py
+++ b/promgen/views.py
@@ -230,6 +230,12 @@ class AuditList(LoginRequiredMixin, ListView):
                         content_type_id=ContentType.objects.get_for_model(models.Rule).id,
                         object_id__in=obj.rule_set.values_list("id", flat=True),
                     )
+                if key == "service":
+                    # Only services may have projects
+                    qset |= Q(
+                        content_type_id=ContentType.objects.get_for_model(models.Project).id,
+                        object_id__in=obj.project_set.values_list("id", flat=True),
+                    )
                 if key == "project":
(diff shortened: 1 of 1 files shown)


promgen/models.py
```diff
@@ -210,6 +210,7 @@ class Service(models.Model):
     )
 
     notifiers = GenericRelation(Sender)
+    project_set = GenericRelation("Project")
```
kfdm, line 213:
```suggestion
```

I believe this is not actually needed. This is already defined by Django automatically due to the Project model having a foreign key to Service

https://github.com/line/promgen/blob/0dc687f1487ed627f149bc45c060fb3f20162b48/promgen/models.py#L244

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey.related_name

GenericRelation is probably incorrect since GenericRelation is supposed to be used with GenericForeignKey which is why it is used with Senders and Rules

https://github.com/line/promgen/blob/0dc687f1487ed627f149bc45c060fb3f20162b48/promgen/models.py#L416
https://github.com/line/promgen/blob/0dc687f1487ed627f149bc45c060fb3f20162b48/promgen/models.py#L82

vincent-olivert-riera: True 👍🏻 I have removed that part.

kfdm (approved):
LGTM

PR #473, opened 2024-01-24
signals.py: Add notifiers to audit logs

vincent-olivert-riera:
We had a query for listing notifiers in views.py::AuditList, however, we never registered any logs from signals.py, therefore the result of that query was always empty.

This commit adds the necessary instructions for registering logs when creating, updating or deleting notifiers, so they will appear in the audit logs.


kfdm:
I think we can't just do this as-is, because Sender can sometimes include sensitive information.

https://github.com/line/promgen/blob/f6a00efbca14b7624efd4db436fcb939a72381f3/promgen/models.py#L564-L579

Audit.log() uses model_to_dict to dump all the information, and we need to ensure the `value` field can not accidentally be logged.

vincent-olivert-riera:
> I think we can't just do this as-is, because Sender can sometimes include sensitive information.
> 
> https://github.com/line/promgen/blob/f6a00efbca14b7624efd4db436fcb939a72381f3/promgen/models.py#L564-L579
> 
> Audit.log() uses model_to_dict to dump all the information, and we need to ensure the `value` field can not accidentally be logged.

According to my tests, when the "alias" field has been populated, that is what is displayed in the logs. Otherwise the value is displayed.

Is that not good enough? If users do not set an alias, I guess we can assume they don't care if the value is displayed.

kfdm:
Try clicking the 'show' button to see all the details that are stored 
スクリーンショット 2024-01-24 16 28 59

vincent-olivert-riera:
> Try clicking the 'show' button to see all the details that are stored スクリーンショット 2024-01-24 16 28 59

Oh, I didn't check that "Show" button. Thanks!

I fixed that issue in [5df5890](https://github.com/line/promgen/pull/473/commits/5df5890c74f356fe1f4b2e66c1b145d1d154afa7).

kfdm:
LGTM. Rebase and I'll do a final check.

vincent-olivert-riera:
> LGTM. Rebase and I'll do a final check.

Done.

kfdm (approved):

PR #474, opened 2024-01-26
Retries for notifications

vincent-olivert-riera:


vincent-olivert-riera:
Need to check why the tests are failing.

PR #482, opened 2024-02-19
[IMPROVEMENT] Rewrite copyto in plain javascript

kfdm:
As we work towards removing our dependence on jQuery, we want to rewrite our copyto helper in plain javascript

### DIFF
--- a/promgen/static/js/promgen.js
+++ b/promgen/static/js/promgen.js
@@ -31,6 +31,22 @@ function update_page(data) {
   }
 }
 
+// Wait until DOM is loaded to add our extra listeners
+document.addEventListener("DOMContentLoaded", function () {
+  /*
+    Copy to shortcut
+
+    Example: <code data-copyto="#id_duration">30s</code>
+  */
+  document.querySelectorAll("[data-copyto]").forEach(srcElement => {
+    srcElement.style.cursor = "pointer";
+    srcElement.addEventListener("click", e => {
+      const dstElement = document.querySelector(srcElement.dataset.copyto);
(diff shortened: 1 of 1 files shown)


promgen/static/js/promgen.js
```diff
@@ -31,6 +31,22 @@ function update_page(data) {
...
+document.addEventListener("DOMContentLoaded", function () {
+  /*
+    Copy to shortcut
+
+    Example: <code data-copyto="#id_duration">30s</code>
+  */
+  document.querySelectorAll("[data-copyto]").forEach(srcElement => {
+    srcElement.style.cursor = "pointer";
+    srcElement.addEventListener("click", e => {
+      var dstElement = document.querySelector(srcElement.dataset.copyto);
```
vincent-olivert-riera, line 44: I always recommend using `const` by default, unless the variable needs to be reassigned, in which case I recommend using `let`. In my opinion `var` should always be avoided.

That is my opinion on JS best practices, but you may have others, hence this comment is optional.

vincent-olivert-riera (approved):

promgen/static/js/promgen.js
kfdm: Sounds reasonable. I think I just don't have enough awareness of javascript best practices since I don't write it as much. Changed to const but just force-pushed for the single, small change.

vincent-olivert-riera (approved):

promgen/static/js/promgen.js
vincent-olivert-riera: Re-approved :+1:

PR #498, opened 2024-04-22
[BUGFIX] Use additional v-pre tags to avoid issues with user input.

kfdm:
We want to add v-pre automatically to all text area tags, since those are often used for general user descriptions.
We also add v-pre tags for other areas where an objects .description tag might be rendered.


sh-kawakami (approved):

vincent-olivert-riera (approved):
LGTM

PR #509, opened 2024-05-15
Update docker-compose.yml

shiresn:
according to mysql 8.4 release note,  The deprecated mysql_native_password authentication plugin is now disabled by default. It can be enabled by starting MySQL with the new --mysql-native-password=ON server option.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,7 +76,7 @@ services:
 
   mysql:
     image: mysql:8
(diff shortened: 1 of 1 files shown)


kfdm (approved):
LGTM. I think I'll pin mysql to 8.4 in a separate PR, just to be more explicit, but the change itself seems reasonable :)

PR #515, opened 2024-05-24
Django fixes

kfdm:
In #495 ( 06a2c20c2a09dfd284d1cb16a4786eedb85e99fb ) we upgraded Django. This means that Django itself doesn't depend on pytz, but we still had a few instances of our code requiring it. Instead of making pytz a dependency of Promgen, we can refactor a few bits to remove the dependency.


vincent-olivert-riera (approved):
LGTM

PR #516, opened 2024-06-03
Fix promql-query Vue component

vincent-olivert-riera:
This component had some issues that prevented it to render properly.

First, the unnecessary usage of 'this_ = this'. In a Vue component, 'this' always refers to the component itself.

Second, the unreliable way to change the display CSS property by accessing 'this_.$el.style.display'. In Vue, we can do conditional rendering using the v-show directive. We have defined a new 'ready' variable that becomes 'true' when the component is ready to be rendered, and use v-show to toggle the visibility of the component based on the value of that variable.

Third, the position of the '<slot></slot>' in the template. This slot is meant to allow the user to describe the result of the query. For instance, if the result of the query is the number of samples, a good content of that slot would be the string 'Samples:', in order to display (for instance) 'Samples: 238' on the screen. However, since the slot was located at the right-hand side of the query result, this would have been displayed instead: '238 Samples:'.

Finally, we have simplified the logic of the 'fetch' call to remove duplicity and reduce the number of lines.

---

**Before:**

image

**After:**

image

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -189,6 +189,7 @@ app.component("promql-query", {
     data: function () {
         return {
             count: 0,
+            ready: false,
         };
     },
     mixins: [mixins],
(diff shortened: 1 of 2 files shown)


kfdm (approved):
LGTM

PR #518, opened 2024-06-04
Shard: Add support for Authorization header

vincent-olivert-riera:
The API end point of some shards are protected with HTTP Basic Auth. This will make the promql-query component to fail receiving a '401 Unauthorized' error response.

To address that, we have added the possibility to specify HTTP Basic Auth credentials on shards.

Apart from that, the query is performed in the back end for security issues. Since the request's "Authorization" header contains the base64-encoded user and password, if we performed the query in the front end that information would be easy to obtain by inspecting the HTTP requests.


promgen/views.py
```diff
@@ -1374,3 +1376,31 @@ def post(self, request, pk):
...
+
+        shard = models.Shard.objects.get(pk=request.GET["shard"])
+        params = {"query": request.GET["query"]}
+        headers = {}
+
+        if shard.basic_auth:
+            user_pass = b64encode(
+                f"{shard.basic_auth_user}:{shard.basic_auth_password}".encode()
+            ).decode()
+            headers["Authorization"] = f"Basic {user_pass}"
```
kfdm, line 1394:
```suggestion
        if shard.basic_auth_user:
            user_pass = b64encode(
                f"{shard.basic_auth_user}:{shard.basic_auth_password}".encode()
            ).decode()
            headers["Authorization"] = f"Basic {user_pass}"
```

Python would like you do a check like that, so having an extra `basic_auth` field seems a little unnecessary, since it's used nowhere else, and not even used in the `admin.py` as a kind of filter 🤔

promgen/views.py
```diff
@@ -1374,3 +1376,31 @@ def post(self, request, pk):
         return JsonResponse(
             {request.POST["target"]: render_to_string("promgen/ajax_clause_check.html", result)}
         )
+
+
+class PromqlQuery(View):
```
kfdm, line 1381: Some of these views we've put in `proxy.py` (since we're proxying over the Prometheus APIs) I'm also wondering that if we add http basic auth support here, then we need to make the rest of our proxy routes aware of the new basic auth settings

kfdm (commented):
If we need to add Auth for this, instead of limiting this to basic_auth, I'm wondering if it would be better to generalize it, and implement it as an optional `Authorization` header that can be added. Then, this would support not only Basic Auth, but also if an upstream server was using some kind of Token or other authentication scheme 🤔

vincent-olivert-riera:
> If we need to add Auth for this, instead of limiting this to basic_auth, I'm wondering if it would be better to generalize it, and implement it as an optional `Authorization` header that can be added. Then, this would support not only Basic Auth, but also if an upstream server was using some kind of Token or other authentication scheme 🤔

Uhm..., yeah, that would be useful. I will change this PR on that direction 👍

promgen/views.py
vincent-olivert-riera: True, all of them should check if authorization is needed.

vincent-olivert-riera:
@kfdm , I have reworked the PR from scratch because I think is easier to review.

It is now ready for review. Thanks.

promgen/urls.py
```diff
@@ -126,6 +126,8 @@
     path("proxy/v1/silences/<silence_id>", csrf_exempt(proxy.ProxyDeleteSilence.as_view()), name="proxy-silence-delete"),
     # Promgen rest API
     path("rest/", include((router.urls, "api"), namespace="api")),
+    # PromQL Query
+    path("promql-query", views.PromqlQuery.as_view(), name="promql-query"),
```
kfdm, line 130: This is more of a 'thinking out loud' comment, but we have a few internal APIs, like this, exporter test, and so on, I wonder if there would be value in making some kind of `internal/` prefix for them, just to help organize them.  There's also some older API cleanup I would like to so, so picking some kind of internal prefix might help us see what have been migrated so far.

promgen/views.py
```diff
@@ -1374,3 +1375,28 @@ def post(self, request, pk):
...
+    def get(self, request):
+        if not all(x in request.GET for x in ["shard", "query"]):
+            return HttpResponse("BAD REQUEST", status=400)
+
+        shard = models.Shard.objects.get(pk=request.GET["shard"])
```
kfdm, line 1385: This is mostly a style thing, 
```suggestion
        shard = get_object_or_404(models.Shard, pk=request.GET["shard"])
```
but since we already have `get__object_or_404` included, this means if for whatever reason, an invalid shard ID was called, this would result in a 404 instead of a 500.

kfdm (approved):
All seems reasonable. I can probably merge as-is without any complaints, but two questions for consideration first. 😄

promgen/views.py
vincent-olivert-riera: Makes sense. Done in 04b06cc. Thanks :+1:

promgen/urls.py
vincent-olivert-riera: > but we have a few internal APIs, like this, exporter test, and so on, I wonder if there would be value in making some kind of internal/ prefix for them, just to help organize them

We could. But since the others are not already organized under the same `internal/` prefix, I guess we could leave this one as it is, and organize all of them in the future.

Or, if you tell me which `path`s on the `urls.py` file you want organized under an `internal/` prefix, I can do that on this PR.

kfdm (approved):
LGTM

vincent-olivert-riera:
> LGTM

Thanks. Rebased on `master` with fixup commits squashed.

PR #520, opened 2024-06-14
Organize Service View Page with Tabs

bai1024:
Categorizing the service view page into tabs aims to provide consistency and improve the readability of the page.
Remark:  
- The Member tab will be added once that page is implemented.
- The project description will be updated until the design is finalized.

Test at local:
image

### DIFF
--- a/promgen/locale/ja/LC_MESSAGES/django.po
+++ b/promgen/locale/ja/LC_MESSAGES/django.po
@@ -172,3 +172,7 @@ msgstr "Serviceを登録"
 #: templates/promgen/project_detail_hosts.html:49
 msgid "Silence selected hosts"
 msgstr "選択したHostをSilence"
+
+#: templates/promgen/service_detail.html:23
+msgid "Actions"
+msgstr "Actions"
--- a/promgen/templates/promgen/service_action_button_group.html
+++ b/promgen/templates/promgen/service_action_button_group.html
@@ -1,0 +1,44 @@
+{% load i18n %}
+{% load promgen %}
+
+<div style="display: inline-block;" class="service-action-button-group">
+  <div class="btn-group btn-group-sm" role="group" aria-label="...">
+    <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      {% translate "Actions" %} <span class="caret"></span>
+    </button>
+    <ul class="dropdown-menu">
+
+      <li>
+        <form action="{% url 'service-notifier' service.id %}" style="display:inline" method="post" v-pre>{% csrf_token %}
+          <input type="hidden" name="sender" value="promgen.notification.user">
+          <input type="hidden" name="value" value="{{request.user.username}}" />
+          <button>{% translate "Subscribe to Notifications" %}</button>
+        </form>
+      </li>
+
+      <hr>
+
+      <li role="presentation"><a href="{% urlqs 'audit-list' service=service.id %}">{% translate "Edit History" %}</a></li>
+      <li role="presentation"><a href="{% urlqs 'alert-list' service=service.name %}">{% translate "Alert History" %}</a></li>
+
+      <hr>
+
+      <li role="presentation"><a href="{% url 'api:service-rules' name=service.name %}">{% translate "Export Rules" %}</a></li>
+      <li role="presentation"><a href="{% url 'api:service-targets' name=service.name %}">{% translate "Export Service" %}</a></li>
+
+      <hr>
+
+      <li role="presentation">
+        <form method="post" action="{% url 'service-delete' service.id %}" onsubmit="return confirm('{% translate "Delete this service?" %}')" style="display: inline">
+          {% csrf_token %}   
+          <button type="submit" style="color:#d9534f;">{% translate "Delete Service" %}</button>  
+        </form>
+      </li>
+
+    </ul>
+  </div>
+  <a @click="setSilenceDataset" data-service="{{service.name}}" class="btn btn-warning btn-sm ml-2 mr-2">{% translate "Silence" %}</a>
+
+  <a href="{% url 'service-update' service.id %}" class="btn btn-warning btn-sm">{% translate "Edit Service" %}</a>
+
+</div>
--- a/promgen/templates/promgen/service_block_panel_notifiers.inc.html
+++ b/promgen/templates/promgen/service_block_panel_notifiers.inc.html
@@ -1,0 +1,7 @@
+{% load promgen %}
+
+{% if service.notifiers.count %} 
+<div class="panel panel-default">
+  {% include "promgen/notifier_block.html" with object=service show_edit=1 %}
+</div>
+{% endif %}
--- a/promgen/templates/promgen/service_block_panel_rules.inc.html
+++ b/promgen/templates/promgen/service_block_panel_rules.inc.html
@@ -1,0 +1,17 @@
+{% load i18n %}
+{% load promgen %}
+
+{% if service.rule_set.count or request.site.rule_set.count %}
+<div class="panel panel-default">
+  <table class="table table-bordered table-condensed">
+    {% include "promgen/rule_header.html" %}
+    {% include "promgen/rule_block.html" with rule_list=service.rule_set.all toggle=1 delete=1 %}
+    {% include "promgen/rule_block.html" with rule_list=request.site.rule_set.all collapse="service-site-rules-" overwrite_type='service' overwrite_id=service.id %}
+  </table>
+  <div class="panel-footer">
+    <a class="btn btn-default btn-xs" role="button" data-toggle="collapse" href=".service-site-rules-{{service.id}}" aria-expanded="false" aria-controls="collapseExample">
+      {% translate 'Show Site Rules' %}
+    </a>
+  </div>
+</div>
+{% endif %}
--- a/promgen/templates/promgen/service_detail.html
+++ b/promgen/templates/promgen/service_detail.html
@@ -8,15 +8,53 @@ Promgen / Service / {{ service.name }}
 
 {% block content %}
 
-<div class="page-header" v-pre>
-  <h1>Service: {{ service.name }}</h1>
-  {% if service.owner %}
-  <p>{% trans 'Contact' %}: {{service.owner.username}}</p>
-  {% endif %}
+<div class="page-header promgen-flex-space-between-center">
+  <div>
+    <h1>Service: {{ service.name }}</h1>
+    {% if service.owner %}
+    <p>{% translate 'Contact' %}: {{service.owner.username}}</p>
+    {% endif %}
+  </div>
+  
+
+  {% include "promgen/service_action_button_group.html" %}
+
 </div>
 
 {% breadcrumb service %}
 
-{% include "promgen/service_block.html" %}
+<ul class="nav nav-tabs mb-5" role="tablist">
+  <li role="presentation" class="active"><a href="#projects" data-toggle="tab">Projects</a></li>
+  <li role="presentation"><a href="#rules" data-toggle="tab">Rules</a></li>
+  <li role="presentation"><a href="#notifiers" data-toggle="tab">Notifiers</a></li>
+</ul>
+
+<div class="well">
+
+  <div class="tab-content">
+    <div role="tabpanel" class="tab-pane active" id="projects">
+      {% include "promgen/service_detail_projects.html" with project_list=project_list|default:service.project_set.all service=service only %}
+    </div>
+
+    <div role="tabpanel" class="tab-pane" id="rules">
+      <div class="text-right mb-4">
+        <a href="{% url 'rule-new' 'service' service.id %}" class="btn btn-primary btn-sm">{% translate "Register Rule" %}</a>
+      </div>
+      
+      {% include "promgen/service_block_panel_rules.inc.html" %}
+
+    </div>
+
+    <div role="tabpanel" class="tab-pane" id="notifiers">
+      <div class="text-right mb-4">
+        <a href="{% url 'service-notifier' service.id %}" class="btn btn-primary btn-sm">{% translate "Register Notifier" %}</a>
+      </div>
+      
+      {% include "promgen/service_block_panel_notifiers.inc.html" %}
+
+    </div>
+  </div>
+
+</div> 
 
 {% endblock %}
--- a/promgen/templates/promgen/service_detail_projects.html
+++ b/promgen/templates/promgen/service_detail_projects.html
@@ -2,16 +2,19 @@
 
 {% regroup project_list|dictsort:"shard_id" by shard as project_shard_list %}
 
+<div class="text-right mb-4">
+    <a href="{% url 'project-new' service.id %}" class="btn btn-primary btn-sm">{% translate "Register Project" %}</a>
+</div>
+
 {% for shard, shard_projects in project_shard_list %}
 <div class="panel panel-default">
     <div class="panel-heading" v-pre>
         Datasource
         <a href="{{shard.get_absolute_url}}">{{shard.name}}</a>
         ( <a href="{{shard.url}}">{{shard.url}}</a> )
-        <a href="{% url 'project-new' service.id %}?shard={{shard.id}}" class="btn btn-primary btn-xs pull-right">{% trans "Register Project" %}</a>
     </div>
 
     {% include 'promgen/service_detail_projects_table.html' with shard_projects=shard_projects service=service %}
-
+    
 </div>
 {% endfor %}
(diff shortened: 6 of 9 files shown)


promgen/templates/promgen/service_detail.html
```diff
@@ -8,15 +8,96 @@
...
-<div class="page-header" v-pre>
-  <h1>Service: {{ service.name }}</h1>
-  {% if service.owner %}
-  <p>{% trans 'Contact' %}: {{service.owner.username}}</p>
-  {% endif %}
+<div class="page-header promgen-flex-space-between-center">
+  <div>
+    <h1>Service: {{ service.name }}</h1>
+    {% if service.owner %}
+    <p>{% trans 'Contact' %}: {{service.owner.username}}</p>
```
kfdm, line 15: I'll review the rest in a bit, but in general we should probably write these out as `translate` now.

https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#std:templatetag-translate

> The trans tag was renamed to translate. The trans tag is still supported as an alias for backwards compatibility.

```suggestion
    <p>{% translate 'Contact' %}: {{service.owner.username}}</p>
```

promgen/locale/ja/LC_MESSAGES/django.po
```diff
@@ -172,3 +172,11 @@ msgstr "Serviceを登録"
 #: templates/promgen/project_detail_hosts.html:49
 msgid "Silence selected hosts"
 msgstr "選択したHostをSilence"
+
+#: templates/promgen/service_detail.html:52
+msgid "Edit"
+msgstr "Serviceを編集"
```
kfdm, line 178: "Edit" is fairly generic, and used by other simple 'edit' buttons, so I'm concerned it will conflict with other edit buttons. We should probably leave this as "Edit Service" in the html templates and it will work with the existing translate strings.

promgen/templates/promgen/service_detail.html
bai1024: Addressed this in [49aabd1](https://github.com/line/promgen/pull/520/commits/49aabd1298e8c90ad15205fdb0afadca98a0f27d)

promgen/locale/ja/LC_MESSAGES/django.po
bai1024: Address this in [589e4d8](https://github.com/line/promgen/pull/520/commits/589e4d86e4e9c6258e14e6413218573179431443)

promgen/templates/promgen/service_detail.html
```diff
@@ -8,15 +8,96 @@
...
+              <button style="background:none;border: none;padding: 0;">{% translate "Delete Service" %}</button>  
+          </form>
+        </a>
+        </li>
+      </ul>
+    </div>
+
+    <a @click="setSilenceDataset" data-service="{{service.name}}" class="btn btn-warning btn-sm ml-2 mr-2">{% translate "Silence" %}</a>
+
+    <a href="{% url 'service-update' service.id %}" class="btn btn-primary btn-sm">{% translate "Edit Service" %}</a>
```
kfdm, line 49:
```suggestion
    <a href="{% url 'service-update' service.id %}" class="btn btn-warning btn-sm">{% translate "Edit Service" %}</a>
```

In the rest of Promgen, we've used `btn-warning` for any editing action, so changing this to `btn-primary` seems strange . . . that said, having the orange silence button next to an orange edit button feels strange. We may want to re-consider our color conventions so that things are kept consistent across the app.

kfdm:
This only seems to update the `/service/<id>` pages, but most of the same templates are used on the `/services/` list pages and the `/` (home) page. 

What probably needs to happen, is to pull out the new changes for the actions dropdown into their own template (similar to the service_header template) and then that can be reused in the various places and cleaned up all at once.

bai1024:
> What probably needs to happen, is to pull out the new changes for the actions dropdown into their own template (similar to the service_header template) and then that can be reused in the various places and cleaned up all at once.

Updated. Now the action button only used in `server/id` page so don't need to update other place.

promgen/templates/promgen/service_detail.html
```diff
@@ -8,15 +8,72 @@
...
+    <a @click="setSilenceDataset" data-service="{{service.name}}" class="btn btn-warning btn-sm ml-2 mr-2">{% translate "Silence" %}</a>
+
+    <a href="{% url 'service-update' service.id %}" class="btn btn-warning btn-sm">{% translate "Edit Service" %}</a>
+
+  </div>
 </div>
 
 {% breadcrumb service %}
 
-{% include "promgen/service_block.html" %}
```
kfdm, line 20: First, should probably rebase against master.
Second, while this improves the detail page some, it still diverges from the home and service page which I think will cause confusion. We should unify them all at the same time.

`/`
home

`/service/`
service-list

`/service/<id>`
service-detail


This does *not* mean simply removing `service_block` from here. It means that `service_detail` probably doesn't need many changes, but that many of the navigation changes should be in `service_block` so that it applies to all the pages.

bai1024: em.. I see. You mean each service_block should have that tabs. The style change should adapt to each service_block?
Based on the description of the requirements, I understand that this tab is only needed on the detail page. 
Do we need further discussion about this requirements?

kfdm: The goal is to improve the UI to be easier to understand by users. Having multiple versions of the same controls is probably not moving towards that goal 😅

sh-kawakami: @kfdm I want to confirm what you want. you want to apply same changes about what to show about service on the top page/service list page, correct?

kfdm: If the decision is to move multiple options under an `Actions` dropdown, then we should apply the *same* change to each page that shows a Service block. Having two different ways to display the same buttons adds confusion to users.

bai1024: @kfdm san.  I think there is some misunderstanding here. And I want to confirm again about this PR.
> The goal is to improve the UI to be easier to understand by users. 

First, I completely agree with this, and this is the purpose of this PR.

Second, I want to confirm one thing that.
> This does not mean simply removing service_block from here. It means that service_detail probably doesn't need many changes, but that many of the navigation changes should be in service_block so that it applies to all the pages.

In my PR, to display the content in the original service_block more effectively on the service_detail page, the service_detail has been refactored and no longer simply uses the service_block component.

According to what you mentioned last time, you want the blocks displaying the service details to be consistent on every page where a block with service details is shown.

Does this mean you want the changes made on the service_detail page to also apply to the service_block component, changing the layout of every service_block component to include tabs?


>If the decision is to move multiple options under an Actions dropdown, then we should apply the same change to each page that shows a Service block. Having two different ways to display the same buttons adds confusion to users.

But according to this, it seems like you only want to unify the changes to the Actions dropdown.

So, I am not quite clear about your specific intention. If possible, please explain it to me in more detail.

At last, I would like to share my thoughts. I believe it is acceptable for the list page and the detail page to be different. This is also a common practice in UI design. I don't think this will confuse users; rather, it helps users visually distinguish the functionality of each page. The list page serves as a preview, while the detail page uses tabs to more clearly display specific content.

kfdm: I have to respectfully disagree. Sure there are times where you might use different controls for each page, but I think in this case, there is less confusion by keeping both the same.

To summarize, if the service editing buttons are going to be collapsed to a drop down action button on one page, they should be collapsed to the same for each page they are used. This reduces visual confusion on the frontend, and avoids duplicate code in the html templates.

bai1024: I updated according to the conclusion we made at yesterday's meeting:  should apply the action buttons in service_block.
Please check.

promgen/templates/promgen/service_detail.html
```diff
@@ -8,15 +8,66 @@
...
+      <div class="panel panel-default">
+        {% include "promgen/notifier_block.html" with object=service show_edit=1 %}
+      </div>
+      {% endif %} 
+    </div>
+  </div>
+
+</div> 
 
 {% endblock %}
```
kfdm, line 73: Adding comment to overall PR, but with some refactoring here, most of this will be shared with service block.

promgen/templates/promgen/service_detail_projects.html
```diff
@@ -0,0 +1,20 @@
...
+    <div class="panel-heading" v-pre>
+        Datasource
+        <a href="{{shard.get_absolute_url}}">{{shard.name}}</a>
+        ( <a href="{{shard.url}}">{{shard.url}}</a> )
+    </div>
+
+    {% include 'promgen/service_detail_projects_table.html' with shard_projects=shard_projects service=service %}
+    
+</div>
+{% endfor %}
```
kfdm, line 20: Seems to be exactly the same as `service_block_projects`

promgen/templates/promgen/service_detail_projects_table.html
```diff
@@ -0,0 +1,27 @@
...
+  </tr>
+  {% endif %}
+
+  {% include 'promgen/project_row.html' %}
+  {% empty %}
+  <tr>
+      <td colspan="4">No Projects?</td>
+  </tr>
+  {% endfor %}
+</table>
```
kfdm, line 27: Can probably just be inline in `service_block_projects` since it's not reused anywhere else.

kfdm (changes requested):
Looking better!!!

But there still seems to be a lot of duplicate between service_block.html and service_detail.html.

What I would recommend.

Make a 'panel' for each section. Something like

```
service_block_panel_projects.inc.html
service_block_panel_rules.inc.html
service_block_panel_notifiers.inc.html
```

Then from both `service_block.html` and `service_detail` you can include the panel. In `service_block.html` they will be in a list, and in `service_detail.html` they'll be under each tab.

I think since this PR has gone through a few cycles, there might be some other duplicates in there. I'll keep checking a bit more.

promgen/templates/promgen/service_action_button_group.html
```diff
@@ -0,0 +1,48 @@
...
+        <a href="">
+          <form action="{% url 'service-notifier' service.id %}" style="display:inline" method="post" v-pre>{% csrf_token %}
+            <input type="hidden" name="sender" value="promgen.notification.user">
+            <input type="hidden" name="value" value="{{request.user.username}}" />
+            <button style="background:none;border: none;padding: 0;">{% translate "Subscribe to Notifications" %}</button>
+          </form>
+        </a>
+      </li>
+
+    </ul>
```
kfdm, line 42: スクリーンショット 2024-07-29 12 07 50

We may want to consider the order.

`Subscribe to Notifications` feels like it could be moved up to the top.
`Delete Service` is a dangerous action. And though it does have a confirmation popup, it feels like we should put it at the very bottom, but also change the color to show it's a dangerous action. (We should probably have it go to a specific `delete confirmation` page instead of *just* the confirmation popup but that's outside the scope of this PR)

promgen/templates/promgen/service_block_panel_notifiers.inc.html
```diff
@@ -0,0 +1,7 @@
+{% load promgen %}
+
+{% if service.notifiers.count %} 
+<div class="panel panel-default">
+  {% include "promgen/notifier_block.html" with object=service show_edit=1 %}
+</div>
+{% endif %}
```
kfdm, line 7: Bit of a nitpick, but most every other file ends with a blank line and since GitHub is showing this as a warning, we should probably be consistent and end each file with a blank line.

promgen/templates/promgen/service_block_panel_rules.inc.html
```diff
@@ -0,0 +1,17 @@
...
+    {% include "promgen/rule_block.html" with rule_list=service.rule_set.all toggle=1 delete=1 %}
+    {% include "promgen/rule_block.html" with rule_list=request.site.rule_set.all collapse="service-site-rules-" overwrite_type='service' overwrite_id=service.id %}
+  </table>
+  <div class="panel-footer">
+    <a class="btn btn-default btn-xs" role="button" data-toggle="collapse" href=".service-site-rules-{{service.id}}" aria-expanded="false" aria-controls="collapseExample">
+      {% translate 'Show Site Rules' %}
+    </a>
+  </div>
+</div>
+{% endif %}
```
kfdm, line 17: insert blank line

promgen/templates/promgen/service_detail_projects_table.html
kfdm, line 27: insert blank line

promgen/templates/promgen/service_detail.html
```diff
@@ -8,15 +8,53 @@
...
 </div>
 
 {% breadcrumb service %}
 
-{% include "promgen/service_block.html" %}
+<ul class="nav nav-tabs mb-5" role="tablist">
+  <li role="presentation" class="active"><a href="#project" data-toggle="tab">Project</a></li>
+  <li role="presentation"><a href="#rule" data-toggle="tab">Rule</a></li>
+  <li role="presentation"><a href="#notifier" data-toggle="tab">Notifier</a></li>
+</ul>
```
kfdm, line 30: I'm wondering if we should use plural form here.

- Projects
- Rules
- Notifiers

We have probably not been particularly consistent across the project.

kfdm (changes requested):
Overall it's looking better, but still a few questions on specific lines.

promgen/templates/promgen/service_detail.html
bai1024: I agree we use plural form here. I notice in the page hear, we also use plural form such as `Services` `Rules` as title

promgen/templates/promgen/service_action_button_group.html
bai1024: Ok. I update the order and change the color for `Delete Service`

promgen/templates/promgen/service_action_button_group.html
```diff
@@ -0,0 +1,49 @@
...
+        <a href="">
+          <form method="post" action="{% url 'service-delete' service.id %}" onsubmit="return confirm('{% translate "Delete this service?" %}')" style="display: inline">
+            {% csrf_token %}   
+              <button style="background:none;border:none;padding:0;color:#d9534f" >{% translate "Delete Service" %}</button>  
+          </form>
+        </a>
+      </li>
+
+    </ul>
+  </div>
```
kfdm, line 44: Even though `<hr>` is more semantic, I think using bootstrap's `<li role="separator" class="divider"></li>` will result in a cleaner layout

Before
before

After
after

promgen/templates/promgen/service_action_button_group.html
```diff
@@ -0,0 +1,49 @@
...
+    <ul class="dropdown-menu">
+
+      <li>
+        <a href="">
+          <form action="{% url 'service-notifier' service.id %}" style="display:inline" method="post" v-pre>{% csrf_token %}
+            <input type="hidden" name="sender" value="promgen.notification.user">
+            <input type="hidden" name="value" value="{{request.user.username}}" />
+            <button style="background:none;border:none;padding:0;">{% translate "Subscribe to Notifications" %}</button>
+          </form>
+        </a>
```
kfdm, line 19: Wrapping a form with an anchor tag feels hacky, especially with the href set this way (should probably be `href="#"` if it's a 'null' link. Even then, the button inline styles also feel hacky. What if we remove that and move it to a CSS rule ?

```css
.dropdown-menu > li > form > button {
  padding: 3px 20px;
  white-space: nowrap;
  background: none;
  border: none;
}
```

This was me figuring out the minimum to copy from the bootstrap 3 rules.

```suggestion
          <form action="{% url 'service-notifier' service.id %}" style="display:inline" method="post" v-pre>{% csrf_token %}
            <input type="hidden" name="sender" value="promgen.notification.user">
            <input type="hidden" name="value" value="{{request.user.username}}" />
            <button>{% translate "Subscribe to Notifications" %}</button>
          </form>
```

Or we can add a custom class to get around things.

kfdm (commented):
Still reviewing, but two things I'm curious about.

promgen/templates/promgen/service_action_button_group.html
bai1024: image
As you can see in the image attached to the PR description, I have already set the hr tag style, and the original divider style should look like this. . But somehow it didn't work at your local..
I update the hr style again

kfdm:
I've opened up #544 as an attempt to split this into two different PRs. Combining the tab changes and the action button changes in the same PR, made things too difficult to review each time.

In the future, we should try to keep one change to one PR. Also it would help to leave the extra commits around and do a rebase/forcepush when ready for final review. Too many force pushes makes it much harder to review just the changes in isolation.

PR #526, opened 2024-07-03
[IMPROVEMENT] Test metrics validity when testing Exporter

kfdm:
When using the 'test' button for an exporter, we previously *only* checked the status code and assumed things were good. It is very possible for an incorrect endpoint to be registered, so we should ensure that the registered endpoint produces correct Prometheus metrics.

Right now, we will return a somewhat *developer* like output, but I expect to make the output more user friendly as we cleanup some of our internal test URLs.

test exporter

### DIFF
--- a/promgen/views.py
+++ b/promgen/views.py
@@ -11,6 +11,8 @@ import time
 from itertools import chain
 
 import prometheus_client
+from prometheus_client.parser import text_string_to_metric_families
+
 import requests
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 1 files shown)


vincent-olivert-riera (approved):
Code changes look good to me.
Tested in my local env and working fine.

Approved 👍

vincent-olivert-riera:
@kfdm ⬇️ 
> license/cla Expected — Waiting for status to be reported

kfdm:
> @kfdm ⬇️
> 
> > license/cla Expected — Waiting for status to be reported

Yeah, was trying to see why that's stuck, but I'm a bit of a loss. I'm not actually the one that configured that either 🤔

vincent-olivert-riera:
> > @kfdm ⬇️
> > > license/cla Expected — Waiting for status to be reported
> 
> Yeah, was trying to see why that's stuck, but I'm a bit of a loss. I'm not actually the one that configured that either 🤔

Me neither, since I don't even have access to the repo's settings.
Do you have anything left to sign in here? https://cla-assistant.io/line/promgen

PR #527, opened 2024-07-03
[INTERNAL] Move proxy_error to util.proxy_error for reuse

kfdm:
We can move our proxy_error wrapper so that it can be reused in our PromqlQuery View to keep things consistent.


vincent-olivert-riera (approved):
Looks good to me 👍

PR #533, opened 2024-07-25
Linting across the project

kfdm:
Minor linting I noticed while doing some other PRs.


promgen/migrations/0023_shard_authorization.py
```diff
@@ -1,6 +1,5 @@
 # Generated by Django 4.2.11 on 2024-06-28 09:57
 
-import django.core.validators
```
kfdm, line 3: I was originally going to add this to `.git-blame-ignores` but since removing an import or changing `!=` to `is not` *could* be a code change, I decided to leave it in the git blame history.

vincent-olivert-riera (approved):
Looks good to me 👍

PR #534, opened 2024-07-25
Renderer tests

kfdm:
In the future, I would like to improve the way we handle some of the rendering, so the first step is to ensure we have some proper test cases to check for when rendering changes.
    
Most of this is updates for test cases, but also add a minor new route as part of a future cleanup of the public api routes.

### DIFF
--- a/promgen/templates/promgen/navbar.html
+++ b/promgen/templates/promgen/navbar.html
@@ -34,7 +34,7 @@
                         <li role="separator" class="divider"></li>
 
                         <li><a href="{% url 'api:api-root' %}">API</a></li>
-                        <li><a href="{% url 'config-targets' %}">Export Targets</a></li>
+                        <li><a href="{% url 'api:all-targets' %}">Export Targets</a></li>
                         <li><a href="{% url 'api:all-rules' %}">Export Rules</a></li>
                         <li><a href="{% url 'config-urls' %}">Export URL</a></li>
                     </ul>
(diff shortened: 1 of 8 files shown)


promgen/templates/promgen/navbar.html
```diff
@@ -36,7 +36,7 @@
                         <li><a href="{% url 'api:api-root' %}">API</a></li>
                         <li><a href="{% url 'config-targets' %}">Export Targets</a></li>
                         <li><a href="{% url 'api:all-rules' %}">Export Rules</a></li>
-                        <li><a href="{% url 'config-urls' %}">Export URL</a></li>
+                        <li><a href="{% url 'api:all-targets' %}">Export URL</a></li>
```
vincent-olivert-riera, line 39: I have had time to test this yet, but just looking at the changes I feel a little bit confused.

Shouldn't this replace `<li><a href="{% url 'config-targets' %}">Export Targets</a></li>` instead?

kfdm: Yes it should. Goof on my part. I'll fix it in a moment 🙇

kfdm: Since there's only the single line in that file, I've rebased and pushed (instead of a separate commit).
I'll likely do a similar change for config-urls in a separate PR and update the tests there as well.

vincent-olivert-riera (approved):
I tested this in my local environment and looks good to me 👍 

image

PR #535, opened 2024-07-26
Convert silence_row into a Vue component

vincent-olivert-riera:
This will help with future development around the Silence area.

The component allows choosing the color of the matchers, and also emits an event when a matcher is clicked. That way we can reuse the component for multiple purposes, displaying the matchers in a different color and doing an different action when a matcher is clicked.

A few bits had to be added in order to maintain the current style, such as the template and tbody tags.


promgen/templates/promgen/vue/silence_row.html
```diff
@@ -0,0 +1,28 @@
...
+  <td class="col-xs-1">[[ time(silence.startsAt) ]]</td>
+  <td class="col-xs-1">[[ time(silence.endsAt) ]]</td>
+  <td class="col-xs-3 text-wrap">
+    <ul class="list-inline">
+      <li v-for="matcher in silence.matchers">
+        <a
+          @click="$emit('matcherClick', matcher.name, matcher.value)"
+          class="label"
+          :class="`label-${labelColor}`"
+        >
```
kfdm, line 12: This changed from `<a @click="addSilenceLabel(match.name, match.value)" class="label label-warning">` but I'm not seeing anything listening to `matcherClick` yet.

vincent-olivert-riera: The important change is from `@click="addSilenceLabel(match.name, match.value)"` to `@click="$emit('matcherClick', matcher.name, matcher.value)"`.

In the first case (before), we just call that function when the matcher is clicked.
In the second case (now), a event is emitted when the matcher is clicked.

The user of that new component is the one that needs to listen to that event. For instance: `<silence-row @matcher-click="myFunctionCall">`.

You can see that in the places where that component is used:
- https://github.com/line/promgen/pull/535/files#diff-b3a6b004644656acccd8ecfb384f505b5f82d6a55ad9659f7063cbde30c5059bR20
- https://github.com/line/promgen/pull/535/files#diff-bcb5da7dbe068f674aa92ec8e3bd58ae02d0e57a970c654edbfbd5857386a427R50
- https://github.com/line/promgen/pull/535/files#diff-89073809b49362f0be626091d8cf08c1ef24e952cfaa0e3852a57c5b77824ba4R22

kfdm: Ok, I guess it threw me off where the event is emitted as `matcherClick` but other places watch for `@matcher-click`. I guess I need to see the Vue documentation to understand how it maps camel case to kebab case

vincent-olivert-riera: Yeah, it's exactly that, the camelCase and kebab-case thing :)

kfdm: So what I keep going back and forth on, I'm not yet sure I understand where making a custom event improves things. 
We talked about handling clicking on matchers slightly different between the silence list and the alert list, but both of those would use different templates apart from `<silence-row>` so there's not really any additional reuse that is added by  this. I think leaving it as `@click` keeps things easier to understand (though I'm still going back and forth on it)

vincent-olivert-riera: > so there's not really any additional reuse that is added by this

Not now, but there will be.

Right now, the `warning` color in the `Silence` button means "create a silence". For consistency, the `warning` color in a matcher means that a silence with that matcher will be created.

However, in the new modal that is being developed to display a list of silences, there will be a filter feature.
We talked about the behavior for when a matcher is clicked, and we agreed that instead of creating a new silence, what will happen is that the matcher will be added to the filter list.

So, in order to visually represent the difference in behavior for when those matchers are clicked, the idea is to use a different color than `warning` (which is used for creating silences).

kfdm: I had already forgotten the filter menu we talked about. In makes more sense in that perspective 🙇

kfdm (approved):

promgen/templates/promgen/vue/silence_row.html
vincent-olivert-riera: No problem :)

PR #539, opened 2024-08-20
Implement "Silence List" Feature as a Modal

bai1024:
1. Change the current list interface to a modal design.
2. Within the modal, Provide table about Silence list
3. Introduce filter feature within the modal interface.

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -167,8 +167,8 @@ app.component("silence-row", {
     },
 });
 
-app.component('silence-modal', {
-    template: '#silence-modal-template',
+app.component('silence-create-modal', {
+    template: '#silence-create-modal-template',
     delimiters: ['[[', ']]'],
     data: () => ({
         state: silenceStore.state,
@@ -215,7 +215,7 @@ app.component('silence-modal', {
                 });
         },
         hideModal() {
-            const modal = $('#silenceModal');
+            const modal = $('#silenceCreateModal');
             if (modal.length) {
                 globalStore.setMessages([]);
                 this.form = {};
@@ -224,7 +224,7 @@ app.component('silence-modal', {
             }
         },
         showModal() {
-            const modal = $('#silenceModal');
+            const modal = $('#silenceCreateModal');
             if (modal.length) {
                 // Detect when the modal is closed, and update the state accordingly. This is
                 // necessary in case the user closes the modal by clicking outside of it, instead of
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -83,6 +83,9 @@ const app = Vue.createApp({
             this.setSilenceLabels(event.target.dataset);
             this.addSilenceLabel('instance', this.selectedHosts.join('|'));
         },
+        openSilenceListModal() {
+            silenceListStore.showModal();
+        },
         fetchSilences: function () {
             fetch('/proxy/v1/silences')
                 .then(response => response.json())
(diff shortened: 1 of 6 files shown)


promgen/templates/promgen/vue/silence_table_modal.html
```diff
@@ -0,0 +1,61 @@
...
+                <th>Matchers</th>
+                <th>Comment</th>
+                <th>Created by</th>
+                <th>Action</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            <template v-for="silence in filteredSilences" :key="silence.id">
+              <silence-row :silence="silence" label-color="warning" @matcher-click="addFilterLabel" />
```
vincent-olivert-riera, line 53: I will do an in-depth review later since I'm currently busy, but this is something I spotted while having a quick look.

The `warning` color is something we normally use when a button **creates** a silence. In this case, we are **filtering** silences, so please use a different color like `info`, for instance.

```suggestion
              <silence-row :silence="silence" label-color="info" @matcher-click="addFilterLabel" />
```

promgen/static/js/promgen.vue.js
```diff
@@ -323,3 +351,95 @@ app.component('exporter-test', {
...
+                label: '',
+                value: ''
+            },
+            store: dataStore
+        };
+    },
+    computed: {
+        globalMessages() {
+            return globalStore.state.messages;
+        },
```
kfdm, line 373: I think this is not being used anywhere?

promgen/static/js/promgen.vue.js
```diff
@@ -323,3 +351,95 @@ app.component('exporter-test', {
...
+        activeSilences() {
+            return this.$root.activeSilences || [];
+        },
```
kfdm, line 376: Since we can get a list from activeSilences() it seems like we could do a quick filter on that, to provide a list of suggestions for the matcher label field, and then do something similar for the matcher value based on which label is selected 🤔

kfdm (commented):
First review pass, but mostly some questions about implementation due to my own ignorance.

promgen/static/js/promgen.vue.js
```diff
@@ -323,3 +351,95 @@ app.component('exporter-test', {
...
+        showModal() {
+            const modal = $('#silenceTableModal');
+            if (modal.length) {
+                modal.on('hidden.bs.modal', () => {
+                    silenceTableStore.hideModal();
+                });
+                modal.modal('show');
+            }
+        },
```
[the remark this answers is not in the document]
bai1024: In this case, $ is indeed jQuery, not something built into Vue. We're using it here because our Modal is still using Bootstrap 3, which relies on jQuery for DOM manipulation and event triggering.

promgen/static/js/promgen.vue.js
bai1024: ah, I see. I will look into it.

promgen/static/js/promgen.vue.js
```diff
@@ -323,3 +326,133 @@ app.component('exporter-test', {
...
+        activeSilences() {
+            return this.$root.activeSilences || [];
+        },
+        filteredSilences() {
+            if (!this.state.labels || this.state.labels.length === 0) {
+                return this.activeSilences;
+            }
+
+            return this.activeSilences.filter(silence => {
+                return this.state.labels.some(filterLabel => {
```
vincent-olivert-riera, line 379:
```suggestion
                return this.state.labels.every(filterLabel => {
```

If we have multiple filters, we want all of them to match. This is the same behavior that Alertmanager has.

vincent-olivert-riera (changes requested):
So now instead of being able to write a custom matcher, we offer the user a selection of existing labels and values. That makes it easier since the user does not even need to type, but removes the possibility to write a regex.
Anyway, for the sake of simplicity I am fine with it. If we want, in the future, we can improve that area to support regex. No problem at all :+1: 

- I have made one code suggestion related to the behavior of the filtering, which acts like a logical `OR`: it returns all silences that match with any of the matchers.
We want that filtering mechanism to work in the same way as Alertmanager, which is like a logical `AND`.
The rest of the code is fine by me.

- Regarding Git stuff, please put in separate commit all the changes related to the renaming of `silence-modal` into `silence-create-modal`.

Good job :)

promgen/static/js/promgen.vue.js
bai1024: Thank you. 
And for this one
>Regarding Git stuff, please put in separate commit all the changes related to the renaming of silence-modal into silence-create-modal.

Do I need to make a different PR for that?

vincent-olivert-riera: > Is it better to make a different PR for that?

No. Even if they are unrelated changes, it is totally fine to have a refactor commit in this PR, so your actual changes will fit better.

So, having one commit for the rename, and another commit for all the changes related to the new silence list modal is fine.

bai1024: No problem. Updated.

vincent-olivert-riera: I have tested it in my local environment and works well :+1: 

Nice job :)

vincent-olivert-riera (approved):

PR #543, opened 2024-09-09
Update translate tag

kfdm:
Update translate tag.

Going forward, the translate tag is the recommended version so we update things for clarity.

https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#std:templatetag-translate

> The trans tag was renamed to translate. The trans tag is still supported as an alias for backwards compatibility.

This will also help remove a bit of noise in #520 regarding some lines that only change the tag.


vincent-olivert-riera:
@kfdm , I would like to merge this PR. Could you rebase it on `master` and fix the conflicts?

PR #558, opened 2024-12-12
Handle query for the audit logs of deleted objects

hoangpn:
If an object is **deleted**, the current code cannot find the object or its related objects, and causes an error in the Audit Log feature (Edit History) when we try to filter that object. In that case, we directly use the **object_id** and **content_type_id** fields of the **Audit model** to find it and skip finding the related objects.

**AS-IS:** 
image

image

**TO-BE:**
image

image


vincent-olivert-riera (approved):
Looks good to me.

PR #575, opened 2025-03-11
Display data source's usage percentage

hoangpn:
Currently, when viewing datasource information, users can directly see the number of samples and exporters that have been registered. However, those numbers don't mean much to our users. Users can also view the usage percentage by hovering, but many users forget to do that, or they don't know such functionality exists.
Therefore, we want to clearly display the usage percentage so our users can make better decisions on where to register their projects, and at the same time help us keep our data sources more or less balanced. The information about the number of samples and exporters registered will be changed to display on hover.

**BEFORE**:
Registering/Updating Project:
image
List/Detail Datasource:
image

**AFTER**:
Registering/Updating Project:
image
List/Detail Datasource:
image


promgen/static/js/promgen.vue.js
```diff
@@ -273,6 +273,9 @@ app.component("promql-query", {
             if (this.count == 0) return "label label-default";
             return "label label-success";
         },
+        maxAsInt() {
+            return Number.parseInt(this.max);
+        },
```
vincent-olivert-riera, line 278: It would be nice to reuse this computed property in line 266:

```
        load: function () {
            return this.count / this.maxAsInt;
        },
```

vincent-olivert-riera (changes requested):
Overall looks good. Just one minor request about reusing code.

promgen/static/js/promgen.vue.js
vincent-olivert-riera: I have seen [496229a](https://github.com/line/promgen/pull/575/commits/496229ab94f8469d55fc6097279811e1fbb0c763). Looks good to me.
Please rebase on master and squash all fixup commits for final review.

vincent-olivert-riera (changes requested):
Looks good. Please rebase and squash fixup commits.

vincent-olivert-riera (approved):

PR #581, opened 2025-04-08
Make owner fields not nullable

hoangpn:
Previously, the owner field in Service, Project, Farm, and Sender models was nullable to accommodate legacy data during migration. This change enforces the owner field to be non-nullable, ensuring data consistency.

`owner = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)`

Currently, only the administrator can delete a user. They will get an error about requiring deleting the following protected related objects:

image


promgen/migrations/0026_alter_farm_owner_alter_project_owner_and_more.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("promgen", "0025_farm_owner"),
```
vincent-olivert-riera: This whole file would need to be regenerated. It will change its name and this dependency.

vincent-olivert-riera (approved):

PR #582, opened 2025-04-08
Add 'owner' field to Farm object

hoangpn:
We add an 'owner' field to the Farm object.
By default, the user who creates a Farm will become the owner. Farm's owner could be changed.


promgen/templates/promgen/farm_detail.html
```diff
@@ -4,7 +4,11 @@
 {% block content %}
 
 <div class="page-header" v-pre>
-  <h1>Farm: {{ farm.name }} ({{ farm.source }})</h1>
+  <h1>Farm: {{ farm.name }} ({{ farm.source }})
```
vincent-olivert-riera, line 7:
```suggestion
  <h1>
    Farm: {{ farm.name }} ({{ farm.source }})
```

If the content of the `<h1>` element is going to be split into multiple lines, do it from the very beginning.

promgen/templates/promgen/farm_detail.html
```diff
@@ -4,7 +4,11 @@
...
+    {% if farm.owner %}
+      <small class="pull-right">{% trans 'Contact' %}: {{farm.owner.username}}</small>
```
vincent-olivert-riera, line 9:
```suggestion
      <small class="pull-right">{% trans "Contact" %}: {{farm.owner.username}}</small>
```

Use double quotes as much as possible.

promgen/views.py
```diff
@@ -908,11 +908,18 @@ class FarmRegister(LoginRequiredMixin, FormView, mixins.ProjectMixin):
 
     def form_valid(self, form):
         project = get_object_or_404(models.Project, id=self.kwargs["pk"])
-        farm, _ = models.Farm.objects.get_or_create(source=discovery.FARM_DEFAULT, **form.clean())
+        farm, _ = models.Farm.objects.get_or_create(
+            source=discovery.FARM_DEFAULT,
+            **form.clean(),
+            defaults={"owner": self.request.user},
```
vincent-olivert-riera, line 914:
```suggestion
```

You don't need this line because you already have the `def get_initial` stuff.
In fact, this line is not allowing to set a different owner than the current user when creating the farm.

That `defaults` line would make sense for objects that are created without specifying an owner, defaulting always to the current user. This is the case of the `Sender` objects (the notifiers).

vincent-olivert-riera (changes requested):
As always, thanks for this PR.

I have made code suggestions for some issues that I found. Please have a look.

promgen/views.py
hoangpn: Oh. I got it. Thank you so much. 🙇

promgen/templates/promgen/farm_detail.html
vincent-olivert-riera: Fixed in d176e52.

promgen/templates/promgen/farm_detail.html
vincent-olivert-riera: Fixed in d176e52.

promgen/views.py
vincent-olivert-riera: Fixed in d176e52.

vincent-olivert-riera (changes requested):
Thanks for addressing my comments. It looks now good for final rebase with fixup commits squashed.

vincent-olivert-riera (approved):
LGTM

PR #583, opened 2025-04-14
Make the Farm retrieve APIs return the name of the Farm's owner

hoangpn:
Field 'owner' was added to the Farm model in commit 2dd4338. The owner's name should be returned through the Farm retrieve APIs, instead of the owner's primary key. This was overlooked on a previous pull request. (https://github.com/line/promgen/pull/582)

AS-IS:
image

TO-BE:
image


vincent-olivert-riera (approved):
LGTM

PR #588, opened 2025-05-14
[BUGFIX] "Home" link in the breadcrumb redirects to the "Services" page

hoangpn:
On some pages of Promgen, the Home link in the breadcrumb is redirected to the Services page instead of the Home page. We have fixed all these errors by using the "breadcrumb" template tag. We also made some improvements to this template tag to handle the Farm object and add the "v-pre" tag by default.

### DIFF
--- a/promgen/templates/promgen/farm_detail.html
+++ b/promgen/templates/promgen/farm_detail.html
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% load i18n %}
+{% load promgen %}
 
 {% block content %}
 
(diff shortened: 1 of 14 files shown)


vincent-olivert-riera (changes requested):
Hello @hoangpn,

Your single commit is doing too many things. Basically, what you did is:
- Add the `v-pre` tag to the `breadcrumb` template tag
- Add the `farm` generator to the `breadcrumb` template tag 
- Reuse the `breadcrumb` template tag everywhere

Could you please split your changes in 3 separate commits?
You can use my list above for the commit message titles, and use your own words to explain the reason (and other details) on each commit message body.

PR #593, opened 2025-05-22
Trim whitespaces of search input

hoangpn:
Sometimes, when users want to search for a keyword, it may accidentally include leading or trailing spaces because the user copied the keyword from elsewhere or due to a typing error. This results in empty search results because nothing matches the input. Therefore, before starting the search, we clean up the user's input by trimming the spaces at the beginning and at the end of the input string.

### DIFF
--- a/promgen/views.py
+++ b/promgen/views.py
@@ -1211,6 +1211,9 @@ class Search(LoginRequiredMixin, View):
         # To avoid searching all object, remove empty search parameters
         # from the request query string.
         query_dict = request.GET.copy()
+        for key, value in query_dict.items():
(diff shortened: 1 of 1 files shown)


hoangpn:
@vincent-olivert-riera
Implementing the trim function for the frontend doesn't seem straightforward, so I only implemented it on the backend. Is this okay? 🤔

vincent-olivert-riera (approved):

vincent-olivert-riera:
> @vincent-olivert-riera Implementing the trim function for the frontend doesn't seem straightforward, so I only implemented it on the backend. Is this okay? 🤔

Yes, this is only needed in the back end.

PR #594, opened 2025-05-22
Disengage Farm Update from Project

hoangpn:
Farms are standalone objects. They just happen to be linkable to Projects. Therefore, I made some changes on Promgen UI to disengage Farm Update from Project.

### The Project Detail Page:
**Farm editable:**
AS-IS:
image
TO-BE:
image
**Farm non-editable:**
AS-IS:
image
TO-BE:
image

### The Farm Detail Page:
**Farm editable:**
AS-IS:
image
TO-BE:
image

**Farm non-editable:**
AS-IS:
image
TO-BE:
image

### The Farm Register page:
AS-IS:
image
TO-BE:
image

### The Farm Update page:
AS-IS:
image
TO-BE:
image

### DIFF
--- a/promgen/templates/promgen/farm_register.html
+++ b/promgen/templates/promgen/farm_register.html
@@ -7,12 +7,16 @@
   <h1>Project: {{ project.name }}</h1>
 </div>
 
-{% breadcrumb farm view.button_label %}
+{% breadcrumb project view.button_label %}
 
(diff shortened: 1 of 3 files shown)


hoangpn:
I've force pushed to fix a minor bug on UI found when I do self-testing.

PR #597, opened 2025-05-28
Improve the feature displaying data source load

hoangpn:
This PR includes two improvements for the feature that displays the load of data sources.

One of the most important changes is replacing the Vue component and the corresponding API that Promgen Web uses to query metrics from data sources with a new component that has better security, and I also removed the old component.

Additionally, I have made changes to always display the load of the data source, regardless of whether the proxy attribute is True or False.

### DIFF
--- a/promgen/templates/promgen/project_form.html
+++ b/promgen/templates/promgen/project_form.html
@@ -85,17 +85,12 @@
         <td>
           <input type="radio" name="shard" value="{{shard.id}}" {% if shard.id == form.shard.value %}checked{% elif shard.enabled is False %}disabled{% endif %}>
         </td>
-        {% if shard.proxy %}
         <td>
           <data-source-usage class="label" shard="{{shard.name}}" metric="samples" max="{{shard.samples}}"></data-source-usage>
         </td>
         <td>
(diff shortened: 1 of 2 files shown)


vincent-olivert-riera (changes requested):
I found a minor thing related to a missing new-line character. Other than that, the rest of the changes look good.

But I have one two requests.

1 - Please squash commits 8adb2b27c5c7e9c9c45c5107ac1d02ecb5044fca and f352525caa5d28ff96032bec4d0426af60d1bae7. After applying the first commit, the `promql-query` Vue component becomes totally unused, therefore what you are actually doing with that commit is replacing it with a new component. Please modify the commit message accordingly.

2 - As we have talked in private, some times the changes done to the `promgen.vue.js` file are not reflected unless you clear your browser's cache. This is because despite of changing the contents of the file, it keeps having the same name, and the browser uses the one that already have in its cache. This can be easily fixed if we pass a query string when we include the file. For instance, we could modify this (in the `base.html` file):
From this:
```
  <script src="{% static 'js/promgen.vue.js' %}"></script>
```
To this:
```
  <script src="{% static 'js/promgen.vue.js' %}?v=1"></script>
```
And increment that number every time we modify the file.

vincent-olivert-riera (approved):

PR #598, opened 2025-05-29
Rename some buttons

hoangpn:
There are some buttons that represent actions, but the button names are displayed as nouns. This can be ambiguous for users. For example, a button that allows users to view the edit history of an object is named "Edit History," but at first glance, users might misunderstand it as a button related to editing some historical information. This is especially true when the button is placed in a group of "Actions" buttons alongside other buttons that indicate actions, such as "Export", "Subscribe" and "Delete."

Therefore, we have changed the names of some buttons to clarify their meanings.
- Edit History -> View Edit History
- Alert History -> View Alert History

AS-IS:
image

TO-BE:
image

### DIFF
--- a/promgen/templates/promgen/project_detail_configuration.html
+++ b/promgen/templates/promgen/project_detail_configuration.html
@@ -26,8 +26,8 @@
         {% trans "Change History" %} <span class="caret"></span>
       </button>
       <ul class="dropdown-menu">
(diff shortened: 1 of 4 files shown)


vincent-olivert-riera (approved):
Looks good to me.

@hoangpn , could you rebase this branch on master before I merge it?

PR #599, opened 2025-05-30
Set owner when adding external farms

vincent-olivert-riera:
Promgen's Farm system is a pluggable system that allows people to write their own plugins. The default plugin is what allows creating local farms, also known as "promgen" farms, and is shipped with Promgen. But other plugins can be added to discover external farms, and link them to projects. After linking them to a project, if that external farm was not already in Promgen's data base, it will be added.
This commit ensures that when such external farms are added to the data base they have an owner set.
Those external farms are not editable, so the owner fields won't be used to decide who can modify the farm. However, since we are planning to enforce the owner fields to not be null, we need to ensure the owner field is set.

### DIFF
--- a/promgen/views.py
+++ b/promgen/views.py
@@ -635,6 +635,9 @@ class FarmLink(LoginRequiredMixin, View):
             logger.info("Importing %s from %s", farm.name, source)
             farm.refresh()
             messages.info(request, "Refreshed hosts")
+            if source != discovery.FARM_DEFAULT:
+                farm.owner = self.request.user
+                farm.save()
         project.farm = farm
         project.save()
         return HttpResponseRedirect(reverse("project-detail", args=[project.id]))


promgen/views.py
```diff
@@ -635,6 +635,9 @@ def post(self, request, pk, source):
             logger.info("Importing %s from %s", farm.name, source)
             farm.refresh()
             messages.info(request, "Refreshed hosts")
+            if source != discovery.FARM_DEFAULT:
+                farm.owner = self.request.user
```
lsegawa, line 639: Wouldn't this reassign the user even if the owner field is already set? 
If the purpose is to ensure owner field to be non-null, then I don't think that modification is needed. Could the conditional perhaps check if the owner field is null?

vincent-olivert-riera: > Wouldn't this reassign the user even if the owner field is already set?

No, because this is inside an higher level `if created:` block. See line [634](https://github.com/line/promgen/pull/599/files#diff-6a13b7546fe1afc70f36c1c1696e55b1d573a559fb49701596187d219943591aR634).
So the code only runs if a new Farm is created after is linked for the first time.

lsegawa: Oh, you are right.

lsegawa (approved):
LGTM

sh-kawakami (approved):

PR #603, opened 2025-06-03
Improve UI for Farm Register page

hoangpn:
We made a few changes to the FarmRegister view class and its template to improve the UI of the Farm Register page. The farm_register.html included two tables from FarmForm and HostForm, which caused many display issues. Therefore, we created a new form named FarmRegisterForm that contains all the fields from both forms to resolve all these issues and ensure the template is consistent with other Promgen's forms.

AS-IS:
image

TO-BE:
image

### DIFF
--- a/promgen/forms.py
+++ b/promgen/forms.py
@@ -254,3 +254,22 @@ class HostForm(forms.Form):
         if not hosts:
             raise ValidationError("No valid hosts")
         self.cleaned_data["hosts"] = list(hosts)
+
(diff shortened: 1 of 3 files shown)


hoangpn:
It is fine at this point for just changing the UI and keep using the current backend code. In the future, it would be nice to have just one form for everything, and combine the "Update" and "Host Add" pages into one.
I will close this ticket. 🙇

PR #609, opened 2025-06-05
Python files cleanup

hoangpn:
Use Ruff to format all of Promgen's Python code and remove any unused imports for cleanup.


vincent-olivert-riera (changes requested):
Looks good to me, but could you please add your commit to the list of ignored commits for Git blame? https://github.com/line/promgen/blob/master/.git-blame-ignore-revs

This is because when we do `git blame` on some file, we don't want to see commits that just did re-formats.

.git-blame-ignore-revs
```diff
@@ -8,3 +8,5 @@ af2c352786f952d819e69b928bf6c99dcde5954b
 5821434b42203b1080dcdc480d1ffcf07fe53a90
 # Format with Ruff
 e72aba6d7e5891ac777f533eef6f7520410c9e0b
+# Format and remove unused imports with Ruff
+5b3d0c07b222ed9f01429f330e5f97e9734b1467
```
vincent-olivert-riera: This file does not have a new-line character at the end of it.

vincent-olivert-riera (changes requested):
Looks good, but your editor didn't put a new-line character at the end of one file. Please fix that :)

.git-blame-ignore-revs
hoangpn: I'm so sorry. I missed it again. 😢 
I'll try to research if there is any automatic solution for PyCharm. 🙇

vincent-olivert-riera (approved):

PR #614, opened 2025-06-17
Not provide query results when testing Rule

hoangpn:
From the Rule's edit page, users can write any query and execute with the test button. After that, the results are presented on the screen. The query is executed on all data sources, and for that Promgen provides the basic auth credentials configured in for each data source. This capability led to a potential security issue where users can run any query anonymously.

Therefore, we want to stop providing the query results to the user, and instead, just say if the rule would fire or not.

**Not firing:**
AS-IS:
image
TO-BE:
image

**Firing:**
AS-IS:
image
TO-BE:
image

### DIFF
--- a/promgen/templates/promgen/ajax_clause_check.html
+++ b/promgen/templates/promgen/ajax_clause_check.html
@@ -4,29 +4,14 @@
   <div class="panel-heading">Query Result</div>
   <div class="panel-body">
   <p>Query Duration: {{ duration }}</p>
-  <p><code>{{ query }}</code></p>
-{% if collapse %}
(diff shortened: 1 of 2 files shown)


promgen/templates/promgen/ajax_clause_check.html
```diff
@@ -4,29 +4,16 @@
...
-{% endif %}
+  {% if firing %}
+  <h3>
+    <span class="label label-danger">Firing</span>
+  </h3>
+  {% else %}
+  <h3>
+    <span class="label label-default">Not firing</span>
+  </h3>
+  {% endif %}
```
vincent-olivert-riera: The `h3` tag is common to both parts of the `if`, so please extract it and leave just the `span` tag inside.

vincent-olivert-riera (approved):

PR #618, opened 2025-06-24
Sort the Farm's sources on the Project Detail page

hoangpn:
We sorted the Farm's sources to ensure that Local Farms always appear first, followed by Remote Farms on the UI when users choose a source to link a Farm to a Project.


promgen/views.py
```diff
@@ -471,7 +471,10 @@ class ProjectDetail(LoginRequiredMixin, DetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["sources"] = models.Farm.driver_set()
+        sources = models.Farm.driver_set()
+        # Sort sources to ensure that they always appear in a consistent order on UI
+        sources = sorted(sources, key=lambda source: (source[1].remote, source[0]))
```
vincent-olivert-riera, line 488: This code could difficult to understand, so perhaps we could rewrite the comment in more detail. Something like this:

```
# Sort sources to ensure local ones appear first on the UI.
#
# - Local sources have the `remote` attribute set to False.
# - In Python, "False < True" returns "True".
# - The sorted function uses the < operator to compare items.
# - When comparing tuples, it first compares the first element of the tuple, after that the second
#   one, and so on so forth.
#
# Based on those four premises, by setting the comparison key as a tuple where the first element is
# the remote attribute we ensure all the local farms are in the first positions, and by having the
# source name as the second element we ensure that they are also alphabetically sorted.
```

promgen/views.py
```diff
@@ -472,8 +472,21 @@ class ProjectDetail(LoginRequiredMixin, DetailView):
...
         sources = models.Farm.driver_set()
-        # Sort sources to ensure that they always appear in a consistent order on UI
+
+        # Sort sources to ensure local ones appear first on the UI.
+        #
+        # - Local sources have the `remote` attribute set to False.
+        # - In Python, "False < True" returns "True".
+        # - The sorted function uses the < operator to compare items.
+        # - When comparing tuples, it first compares the first element of the tuple, after that
+        # the second one, and so on so forth.
```
vincent-olivert-riera: Sorry, this was my mistake when I shared this comment with you.

Could you please indent the text of this line a couple of characters more, so the `the` word starts in the same column as the `When` word?

Example:

```
        # - When comparing tuples, it first compares the first element of the tuple, after that
        #   the second one, and so on so forth.
```

vincent-olivert-riera (changes requested):
@hoangpn , excellent.

I made one request to fix a mistake made by me.

Please, address that, and the next time you push, please rebase on master and squash all the fixups.

It will be ready to approve and merge.

PR #620, opened 2025-06-25
Feature/support all operators for silences

hoangpn:
- Support all operators (`=`,`=~`,`!=`,`!~`) for creating & filtering new silence
- Show matchers with operators on the silence list

### Creating silence
AS-IS:
image

TO-BE:
image

### Silence list
AS-IS:
image

TO-BE:
image

### DIFF
--- a/promgen/static/js/promgen.vue.js
+++ b/promgen/static/js/promgen.vue.js
@@ -99,8 +99,8 @@ const app = Vue.createApp({
             this.setSilenceLabels(event.target.dataset);
             this.addSilenceLabel('instance', this.selectedHosts.join('|'));
         },
-        openSilenceListModal(params) {
-            silenceListStore.showModal(params);
+        openSilenceListModal(params, silences, target) {
+            silenceListStore.showModal(params, silences, target);
         },
         fetchSilences: function () {
             fetch('/proxy/v1/silences')
@@ -137,6 +137,17 @@ const app = Vue.createApp({
             let tgt = document.getElementById(target);
             tgt.setAttribute('list', dst + '.' + src);
         },
+        getActiveSilencesForService: function (service) {
+            return getActiveSilences(this.activeSilences, "service", service);
+        },
+        getActiveSilencesForProject: function (project, service) {
+            const silencesMatchProject = getActiveSilences(this.activeSilences, "project", project);
+            const silencesMatchService = getActiveSilences(this.activeSilences, "service", service);
+            return silencesMatchProject.filter(silence =>
+                !silence.matchers.some(matcher => matcher.name === "service") ||
+                silencesMatchService.includes(silence)
+            );
+        },
     },
     computed: {
         activeServiceAlerts: function () {
@@ -147,12 +158,6 @@ const app = Vue.createApp({
         },
         activeRuleAlerts: function () {
             return groupByLabel(this.activeAlerts, 'alertname');
-        },
-        activeServiceSilences: function () {
-            return groupByLabel(this.activeSilences, 'service');
-        },
-        activeProjectSilences: function () {
-            return groupByLabel(this.activeSilences, 'project');
         },
         activeAlerts: function () {
             return this.globalAlerts.filter(alert => alert.status.state === 'active');
@@ -405,7 +410,7 @@ const silenceListStore = Vue.reactive({
             this.state.labels.splice(index, 1);
         }
     },
-    showModal(params=null) {
+    showModal(params=null, silences=null, target=null) {
         // Accept an optional parameter of type object that contains another "matchers" object. This
         // is to allow opening the modal with some matchers already set for filtering.
         //
@@ -420,6 +425,39 @@ const silenceListStore = Vue.reactive({
         ) {
             for (const [label, value] of Object.entries(params.matchers)) {
                 this.addFilterLabel(label, value);
+            }
+        }
+
+        // Accept a list of silences as an optional parameter to show active silences of a specific project
+        // or service, by creating pre-filled matchers based on the matchers with 'project' or 'service' labels
+        // in the input silences.
+        // parameter 'target' must be 'project' or 'service'.
+        if (silences && typeof silences === "object" && Array.isArray(silences)) {
+            for (const silence of silences) {
+                for (const matcher of silence.matchers) {
+                    if ("service" === target) {
+                        if (["service"].includes(matcher.name)) {
+                            this.addFilterLabel(
+                              matcher.name,
+                              matcher.value,
+                              matcher.isEqual
+                              ? matcher.isRegex ? "=~" : "="
+                              : matcher.isRegex ? "!~" : "!="
+                            );
+                        }
+                    }
+                    if ("project" === target) {
+                        if (["project", "service"].includes(matcher.name)) {
+                            this.addFilterLabel(
+                              matcher.name,
+                              matcher.value,
+                              matcher.isEqual
+                              ? matcher.isRegex ? "=~" : "="
+                              : matcher.isRegex ? "!~" : "!="
+                            );
+                        }
+                    }
+                }
             }
         }
 
@@ -460,15 +498,73 @@ app.component('silence-list-modal', {
                 return this.activeSilences;
             }
 
+            // Group matchers in state by label
+            // For example, if we have matchers like:
+            //   [
+            //      {label: 'service', value: 'foo', operator: '='},
+            //      {label: 'service', value: 'bar', operator: '!='},
+            //      {label: 'project', value: 'baz', operator: '=~'},
+            //   ]
+            // We will create a map like:
+            //   {
+            //     'service': [
+            //          {label: 'service', value: 'foo', operator: '='},
+            //          {label: 'service', value: 'bar', operator: '!='},
+            //     ],
+            //     'project': [{label: 'project', value: 'baz', operator: '=~'}]
+            //   }
+            const groups = this.state.labels.reduce((map, item) => {
+                if (item.label) {
+                    if (!map.has(item.label)) {
+                        map.set(item.label, []);
+                    }
+                    map.get(item.label).push(item);
+                }
+                return map;
+            }, new Map());
+
+            // Filter silences based on the matchers in the groups
+            // A silence matches if all of its matchers match at least one of the matchers in
+            // the groups that have the same label.
+            // For example, if we have a groups map like:
+            //   {
+            //     'service': [
+            //          {label: 'service', value: 'foo', operator: '='},
+            //          {label: 'service', value: 'bar', operator: '!='},
+            //     ],
+            //     'project': [{label: 'project', value: 'baz', operator: '=~'}]
+            //   }
+            //
+            // - This silence will match: 'service=foo, project=~baz'
+            // - This silence will match: 'service!=bar, project=~baz'
+            // - This silence will match: 'service=foo'
+            // - This silence will NOT match: 'service=foo, project=~test'
+            //   because the value 'test' for the label 'project' is not in the groups map.
+            // - This silence will NOT match: 'service=test, project=~baz'
+            //   because the value 'test' for the label 'service' is not in the groups map.
+            // - This silence will NOT match: 'service=foo, project!=baz'
+            //   because the operator '!=' for the label 'project' is not in the groups map.
             return this.activeSilences.filter(silence => {
-                return this.state.labels.every(filterLabel => {
-                    return silence.matchers.some(matcher =>
-                        matcher.name === filterLabel.label &&
-                        matcher.value === filterLabel.value &&
-                        matcher.isEqual === ['=', '=~'].includes(filterLabel.operator) &&
-                        matcher.isRegex === ['=~', '!~'].includes(filterLabel.operator)
-                    );
-                });
+                let doesSilenceMatch = false;
+                for (const matcher of silence.matchers) {
+                    const groupsForMatcher = groups.get(matcher.name);
+                    if (groupsForMatcher) {
+                        const matchFound = groupsForMatcher.some(group =>
+                            matcher.name === group.label &&
+                            matcher.value === group.value &&
+                            (
+                                (matcher.isEqual && ["=", "=~"].includes(group.operator)) ||
+                                (!matcher.isEqual && ["!=", "!~"].includes(group.operator))
+                            )
+                        );
+                        if (!matchFound){
+                            return false;
+                        } else {
+                            doesSilenceMatch = true;
+                        }
+                    }
+                }
+                return doesSilenceMatch;
             });
         },
         uniqueLabels() {
--- a/promgen/templates/base.html
+++ b/promgen/templates/base.html
@@ -62,9 +62,9 @@
   <script type="text/x-template" id="silence-list-modal-template">
     {% include 'promgen/vue/silence_list_modal.html' %}
   </script>
-  <script src="{% static 'js/promgen.js' %}"></script>
+  <script src="{% static 'js/promgen.js' %}?v=1"></script>
   <script src="{% static 'js/mixins.vue.js' %}"></script>
-  <script src="{% static 'js/promgen.vue.js' %}?v=4"></script>
+  <script src="{% static 'js/promgen.vue.js' %}?v=5"></script>
   {% block javascript %}{% endblock %}
 
   <datalist style="display:none" id="common.labels">
--- a/promgen/static/js/promgen.js
+++ b/promgen/static/js/promgen.js
@@ -31,6 +31,35 @@ function update_page(data) {
   }
 }
 
+function doesMatcherMatch(matcher, labelName, labelValue) {
+  if (matcher.name !== labelName) {
+    return false;
+  }
+
+  if (matcher.isRegex) {
+    const regex = new RegExp("^(?:" + matcher.value + ")$");
+    const matches = regex.test(labelValue);
+    return matcher.isEqual ? matches : !matches;
+  } else {
+    return matcher.isEqual ?
+      (matcher.value === labelValue) :
+      (matcher.value !== labelValue);
+  }
+}
+
+function getActiveSilences(items, labelName, labelValue) {
+  const activeSilences = [];
+  for (const item of items) {
+    const matches = item.matchers.some(matcher =>
+      doesMatcherMatch(matcher, labelName, labelValue)
+    );
+    if (matches) {
+      activeSilences.push(item);
+    }
+  }
+  return activeSilences;
+}
+
 // https://blog.bitsrc.io/debounce-understand-and-learn-how-to-use-this-essential-javascript-skill-9db0c9afbfc1
 function debounce(func, delay = 250) {
   let timerId;
--- a/promgen/templates/promgen/project_detail.html
+++ b/promgen/templates/promgen/project_detail.html
@@ -52,8 +52,8 @@ Promgen / Project / {{ project.name }}
 
 <a
   v-cloak
-  v-if="activeProjectSilences.has('{{project.name}}')"
-  @click="openSilenceListModal({ 'matchers': { 'project': '{{project.name}}' } })"
+  v-if="getActiveSilencesForProject('{{project.name}}', '{{project.service.name}}').length > 0"
+  @click="openSilenceListModal(null, silences=getActiveSilencesForProject('{{project.name}}', '{{project.service.name}}'), 'project')"
   class="btn btn-warning btn-sm mb-4"
   role="button"
 >
--- a/promgen/templates/promgen/service_block.html
+++ b/promgen/templates/promgen/service_block.html
@@ -22,8 +22,8 @@
 
   <a
     v-cloak
-    v-if="activeServiceSilences.has('{{service.name}}')"
-    @click="openSilenceListModal({ 'matchers': { 'service': '{{service.name}}' } })"
+    v-if="getActiveSilencesForService('{{service.name}}').length > 0"
+    @click="openSilenceListModal(null, silences=getActiveSilencesForService('{{service.name}}'), 'service')"
     class="btn btn-warning btn-sm mb-4"
     role="button"
   >
--- a/promgen/templates/promgen/service_detail.html
+++ b/promgen/templates/promgen/service_detail.html
@@ -51,8 +51,8 @@ Promgen / Service / {{ service.name }}
 
 <a
   v-cloak
-  v-if="activeServiceSilences.has('{{service.name}}')"
-  @click="openSilenceListModal({ 'matchers': { 'service': '{{service.name}}' } })"
+  v-if="getActiveSilencesForService('{{service.name}}').length > 0"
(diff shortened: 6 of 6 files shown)


vincent-olivert-riera:
@hoangpn , commit [99d1ea3](https://github.com/line/promgen/pull/620/commits/99d1ea3b6c0ddf9f82b7c6716b2a13711c914dee) added support to select all the operators when filtering the global silence list.

There are 3 fields in that form: label, operator and value.
They are selected in order, so a user does this:
- select a label
- select an operator
- select a value
Finally the `+` button is clicked and the filter is created, and the list of silences is modified after applying the new filter.

The reason the 3 fields are selected in order is because the next field should be updated depending on the selection made in the previous field. For instance, imagine we have these silences in the sistem:

- `service="foo",project="bar"
- `service="foo",project=~"baz.*"

Now, when I open the global list of silences I will see all the silences (in our example, two).
Then I create a filter by selecting the `project` label. When I do that, the list of available operators should only contain `=` and `=~`.
If I select the `=` operator, the list of available values should only contain `bar`.
If I select the `=~` operator, the list of available values should only contain `baz.*`.

Could you please make the necessary changes to achieve that behavior?

promgen/prometheus.py
```diff
@@ -288,16 +288,21 @@ def silence(*, labels, duration=None, **kwargs):
...
-            "name": name,
-            "value": value,
-            "isEqual": True,  # Right now we only support = and =~
-            "isRegex": True if value.endswith("*") else False,
-        }
-        for name, value in labels.items()
-    ]
+    if "matchers" in kwargs:
+        # We pass matchers as a list if the method is called from ProxySilencesV2
+        kwargs["matchers"] = json.loads(json.dumps(kwargs["matchers"]))
```
vincent-olivert-riera: This looks very strange.

First, you `json.dumps` that variable. After that, you `json.loads` the result. That, in theory should give you the same result as the original variable, since those two operations are the opposite of each other (is like doing +1 and -1, the final result is the same). And finally, you overwrite that same variable with (in theory) the same value that had from the beginning.

I think you are doing it because you want to convert a list of `OrderedDict` objects, to a list of normal `dict` objects.

However, an `OrderedDict` object can be accessed in the same way as a normal `dict` object, so..., did you try leaving that `kwargs["matchers"]` variable untouched?

I did it and it seems to work just fine:

```
    if "matchers" in kwargs:
        # We pass matchers as a list if the method is called from ProxySilencesV2
        pass
```

Or even better, you can remove the positive part of the `if`, and only use the `else` part like this:

```
    # If no matchers are provided, it means the method is called from ProxySilences (V1). In this
    # case, we need to convert labels to matchers.
    #
    # Otherwise, the method is called from ProxySilencesV2 and the matchers are already provided in
    # the right format, so we don't need to touch them.
    if not "matchers" in kwargs:
        kwargs["matchers"] = [
            {
                "name": name,
                "value": value,
                "isEqual": True,  # Right now we only support = and =~
                "isRegex": True if value.endswith("*") else False,
            }
            for name, value in labels.items()
        ]
```

promgen/static/js/promgen.vue.js
```diff
@@ -27,6 +27,11 @@ const silenceStore = Vue.reactive({
     },
     setLabels(labels) {
         this.state.labels = { ...labels };
+        for (const [key, value] of Object.entries(this.state.labels)) {
+            if (!Array.isArray(value)) {
+                this.state.labels[key] = [value, '='];
```
vincent-olivert-riera: Nice, you are trying to keeping backwards compatibility for old silences that didn't have an operator :+1: 

But, remember that if the `value` of the matcher includes a `*` character, we consider that a regex, therefore the operator should be `=~`.

promgen/static/js/promgen.vue.js
```diff
@@ -75,8 +80,8 @@ const app = Vue.createApp({
         setSilenceDataset(event) {
             this.setSilenceLabels(event.target.dataset);
         },
-        addSilenceLabel(label, value) {
-            silenceStore.addLabel(label, value);
+        addSilenceLabel(label, value, operator = '=') {
```
vincent-olivert-riera: I think `operator` should be set to `undefined` by default, and add more logic inside the `silenceStore.addLabel` function to handle that case.

promgen/static/js/promgen.vue.js
```diff
@@ -75,8 +80,8 @@ const app = Vue.createApp({
...
+            silenceStore.addLabel(label, [value, operator]);
```
vincent-olivert-riera, line 95: In the **future** we should probably rework the `silenceStore` so the `silenceStore.addLabel` function supports 3 arguments (`label`, `value`, `operator`), and the matchers are stored into a list of matchers that would have this structure:

```
matchers: [
  {
     label: "foo",
     operator: "=~",
     value: "bar",
  },
  ...
]
```

But for now, storing both the value and the operator in the same variable (as an array) is fine by me.

**However**, that `silenceStore.addLabel` function needs to handle the case when `operator` is `undefined`, because depending on the value the real operator could be either `=` or `=~`. Something like this:

```
const silenceStore = Vue.reactive({
  ...
  addLabel(label, value) {
    if (Array.isArray(value) && value[1] === undefined) {
      if (value[0].includes("*")) {
        value[1] = "=~";
      } else {
        value[1] = "=";
      }
    }
    
    this.state.labels[label] = value;
  }
```

promgen/static/js/promgen.vue.js
```diff
@@ -184,26 +189,42 @@ app.component('silence-create-modal', {
...
         removeLabel(label) {
             delete this.state.labels[label];
         },
         submit() {
+            matchers = [];
+            for (const [label, value] of Object.entries(this.state.labels)) {
+                matchers.push({
+                    name: label,
+                    value: value[0],
+                    isEqual: ['=', '=~'].includes(value[1]) || value === null || value === undefined,
```
vincent-olivert-riera: Regarding this bit:

```
isEqual: ... || value === null || value === undefined,
```

Taking into account that we are inside the `submit` function for creating silences, what are the cases where the `value` variable is `null` or `undefined`?

promgen/static/js/promgen.vue.js
```diff
@@ -184,26 +189,42 @@ app.component('silence-create-modal', {
...
                 duration: this.form.duration,
                 createdBy: this.form.createdBy,
                 comment: this.form.comment
             });
 
-            fetch('/proxy/v1/silences', { method: 'POST', body })
+            const headers = {
+                'Content-Type': 'application/json',
+                'X-CSRFToken': document.querySelector('input[name="csrf_token"]').value,
+            };
```
vincent-olivert-riera, line 258: Why is this necessary for `/proxy/v2/silences` but was not necessary for `/proxy/v1/silences`?

promgen/templates/promgen/vue/silence_row.html
```diff
@@ -15,7 +15,7 @@
           class="label"
           :class="`label-${labelColor}`"
         >
-          [[ matcher.name ]]:[[ matcher.value ]]
+          [[ matcher.name ]]<span v-if="matcher.isEqual && !matcher.isRegex">=</span><span v-if="matcher.isEqual && matcher.isRegex">=~</span><span v-if="!matcher.isEqual && !matcher.isRegex">!=</span><span v-if="!matcher.isEqual && matcher.isRegex">!~</span>[[ matcher.value ]]
```
vincent-olivert-riera: I don't like this.

We should be able to do this instead:

```
[[ matcher.name ]][[ matcher.operator ]][[ matcher.value ]]
```

`matcher.operator` could be a variable that holds the operator, or it could be a computed property (is like a function) that returns the operator.

This comment also applies for the change you are doing to line number 9 on silence_row.html:

```

-        :title="`${matcher.name}:${matcher.value}`"
+        :title="`${matcher.name}${matcher.isEqual ? (matcher.isRegex ? '=~' : '=') : (matcher.isRegex ? '!~' : '!=')}${matcher.value}`"
```

promgen/static/js/promgen.vue.js
```diff
@@ -439,12 +452,32 @@ app.component('silence-list-modal', {
...
+            this.filteredSilences.forEach((silence) => {
+                silence.matchers.forEach((matcher) => {
+                    if (matcher.name === this.form.label) {
+                        const op = matcher.isEqual ?
+                            (matcher.isRegex ? "=~" : "=") : (matcher.isRegex ? "!~" : "!=");
+                        operators.add(op);
+                    }
+                });
+            });
+          return ["=", "=~", "!=", "!~"].filter((op) => operators.has(op));
```
vincent-olivert-riera: This `return` should be indented two more characters.

promgen/templates/base.html
```diff
@@ -63,8 +63,8 @@
     {% include 'promgen/vue/silence_list_modal.html' %}
   </script>
   <script src="{% static 'js/promgen.js' %}"></script>
-  <script src="{% static 'js/mixins.vue.js' %}"></script>
-  <script src="{% static 'js/promgen.vue.js' %}?v=4"></script>
+  <script src="{% static 'js/mixins.vue.js' %}?v=1"></script>
```
vincent-olivert-riera: Commit 80c6afa06a2d704cd8a742efda0c3db32c3112cb didn't modify the `mixins.vue.js` file. I think you wanted to increase the version of `promgen.js` instead.

vincent-olivert-riera (changes requested):
@hoangpn , I have complete the first review spin.

promgen/static/js/promgen.vue.js
hoangpn: @vincent-olivert-riera 
API v1 is a **Django's View class.** When I wrote API v2, I thought it should use a standard library for APIs, so I used **Django REST Framework's APIView class**. 
When I tested this feature for the first time, it returned a 'missing CSRFToken' error. So I simply added the token.

promgen/static/js/promgen.vue.js
hoangpn: @vincent-olivert-riera 
Ah I see. I forgot the =~ case. For example, when you silence selected hosts.
Thank you so much!

promgen/static/js/promgen.vue.js
hoangpn: @vincent-olivert-riera 
There hasn't been cases like that yet, but when using Alertmanager's API, I noticed that the default value of `isEqual` is **True** if the client does not provide the `isEqual` field.
Therefore, I want to implement this so that its behavior is similar to Alertmanager, in case someone in the future somehow does not pass `isEqual`. 🙇

promgen/static/js/promgen.vue.js
vincent-olivert-riera: Understood, thanks :+1:

promgen/static/js/promgen.vue.js
vincent-olivert-riera: > Therefore, I want to implement this so that its behavior is similar to Alertmanager, in case someone in the future somehow **does not pass isEqual**. 🙇

"Does not pass `isEqual`", to what? To the new Promgen's Silences V2 API that you are adding on this PR?

If that's the case, what we should do is making the `isEqual` field optional with a default value set to `True`, not here, but in the API end point schema:

```
class MatcherSerializer(serializers.Serializer):
    ...
    isEqual = serializers.BooleanField(default=True)
```

promgen/static/js/promgen.vue.js
```diff
@@ -186,6 +186,19 @@ app.component("silence-row", {
...
+            // Key: isEqual_isRegex, Value: the operator
+            return new Map([
+                ['true_false', '='],
+                ['true_true', '=~'],
+                ['false_false', '!='],
+                ['false_true', '!~'],
+            ]);
+        },
+
+    },
```
vincent-olivert-riera, line 208: Regarding the commit...

> fixup! Show operators of Silence's matchers on UI
> Create a computed property for mapping isEqual and isRegex to the operator

I realized I made a mistake in my previous suggestion. We cannot use a computed property because we are iterating over the matchers, and the computed property would be for the silence (which has multiple matchers). And also because we cannot pass arguments to a computed property.

My aim for my previous comment was to keep the template as clean as possible, and put all the logic to determine the operator in the component's JavaScript part. So, we can use a method (which accepts arguments) to return the operator for a given matcher: `getOperator(matcher)`.

This is what I meant: https://github.com/vincent-olivert-riera/promgen/commit/72946186ee5f8c1d0d5bfaa658925b9bbc985f9c

Feel free to cherry-pick that commit.

promgen/static/js/promgen.vue.js
hoangpn: Setting default value for isEqual is a nice comment. I've add a fixup commit for that. 🙇 
Actually, I have tried this before. If I don't pass isEqual on the client side, Promgen BE will record it as null. When the value isEqual=null is passed to Alertmanager, it will become True there. So there's no problem.

What I'm concerned about is something else. If someone reuses my code in other UI and, somehow, they do not check the operator exist and do not pass the isEqual value at the time the **submit()** function is called, then this condition will return **False**, and isEqual will accidentally become False. 
`isEqual: ['=', '=~'].includes(value[1])`

promgen/static/js/promgen.vue.js
```diff
@@ -186,18 +186,22 @@ app.component("silence-row", {
...
+                if (matcher.isRegex) {
+                    return "=~";
+                } else {
+                    return "=";
+                }
+            } else {
+                if (matcher.isRegex) {
+                    return "!~";
+                } else {
+                    return "!";
```
vincent-olivert-riera: This should be `return "!=";`

I made a mistake in my commit, sorry :)

promgen/prometheus.py
```diff
@@ -288,16 +288,21 @@ def silence(*, labels, duration=None, **kwargs):
...
+    # case, we need to convert labels to matchers.
+    #
+    # Otherwise, the method is called from ProxySilencesV2 and the matchers are already provided in
+    # the right format, so we don't need to touch them.
+    if not "matchers" in kwargs:
+        kwargs["matchers"] = [
+            {
+                "name": name,
+                "value": value,
+                "isEqual": True,  # Right now we only support = and =~
```
vincent-olivert-riera: We have a big comment above the `if` that explains everything quite well..., but just in case, I would like to be more explicit here.

Could we please change this line to this?

```
"isEqual": True,  # Only = and =~ are supported in ProxySilences (V1)
```

promgen/serializers.py
```diff
@@ -120,3 +121,53 @@ class HostSerializer(serializers.ModelSerializer):
...
+
+    def validate(self, data):
+        # Validation for matchers
+        if "matchers" not in data or not data["matchers"]:
+            raise errors.SilenceError.NOMATCHER.error()
+
+        # Users should not be able to accidentally silence a global rule without
+        # setting some other labels as well.
+        for matcher in data["matchers"]:
+            if matcher["name"] == "alertname" and matcher["isEqual"] == True:
```
vincent-olivert-riera: Regardless of the operator, we should not allow users to apply silences on global rules without specifying a service or project, right?

I think we should remove the `and matcher["isEqual"] == True` part.

Otherwise we would be allowing users to create a silence with the following matcher: `alertname!="something"`

Also, this code does not check if the rule is actually a global rule. You would need to do something like this:

```
rule = models.Rule.objects.get(name=matcher["value"])
if rule.content_type.model == "site":
  # Now you know it's a global rule (we call them site rules, actually)
```

promgen/static/js/promgen.vue.js
```diff
@@ -27,8 +27,24 @@ const silenceStore = Vue.reactive({
     },
     setLabels(labels) {
         this.state.labels = { ...labels };
+        for (const [key, value] of Object.entries(this.state.labels)) {
+            if (!Array.isArray(value)) {
+                if (value.includes("*")) {
+                    this.state.labels[key] = [value, '=~'];
```
vincent-olivert-riera: That `'=~'` can be wrapped in double quotes. There are other places with this kind of thing.

Could you try passing the new JavaScript code you are adding through Prettier? It will fix all those single quotes, missing semi-colons, indentations and other stuff for you.

promgen/static/js/promgen.vue.js
vincent-olivert-riera: As we discussed in Slack, all the `null` and `undefined` comparisons in `isEqual: ... || value === null || value === undefined,` are not necessary.

promgen/serializers.py
```diff
@@ -120,3 +121,53 @@ class HostSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Host
         exclude = ("id", "farm")
+
+
+class MatcherSerializer(serializers.Serializer):
+    name = serializers.CharField()
+    value = serializers.CharField()
+    isRegex = serializers.BooleanField()
+    isEqual = serializers.BooleanField(required=False, default=True)
```
vincent-olivert-riera: `required` is automatically `False` if you define a `default` value, so this is enough:

```
    isEqual = serializers.BooleanField(default=True)
```

promgen/templates/promgen/vue/silence_create_modal.html
```diff
@@ -56,10 +56,10 @@ <h5>
                   <li v-for="(value, label) in state.labels">
                     <div class="promgen-label-target">
                       <span
-                        :title="`${label}:${value}`"
+                        :title="`${label}${value[1]}${value[0]}`"
                         class="promgen-silence-matcher-truncate"
                       >
-                        [[label]]=[[value]]
+                        [[`${label}${value[1]}${value[0]}`]]
```
vincent-olivert-riera: In the `silence_list_modal.html` file, line number 63, we are doing this:

```
[[item.label]][[item.operator]][[item.value]]
```

What is the reason for using the backtick (`) operator to create a template literal in here? Is it because this code didn't work?

```
[[ label ]][[ value[1] ]][[ value[0] ]]
```

Maybe there is a problem with the brackets :shrug: Could you confirm?

promgen/prometheus.py
vincent-olivert-riera: Fixed in 5913ade11f06a2e0f969c92ef2e6b3db216dd77f

promgen/serializers.py
vincent-olivert-riera: Fixed in 5913ade11f06a2e0f969c92ef2e6b3db216dd77f

promgen/static/js/promgen.vue.js
vincent-olivert-riera: Fixed in 0b14376bbd7f5284e19e6551c3b879fc443bfb0a and 02817bae39e859f915a6e78aee0e4829b62bccf8 and 2ee7eddb0a358d0f774c18ea82f0825dff1376f5.

promgen/templates/promgen/vue/silence_create_modal.html
vincent-olivert-riera: Fixed in 0b14376bbd7f5284e19e6551c3b879fc443bfb0a.

promgen/static/js/promgen.vue.js
vincent-olivert-riera: Fixed in 608c0f8b2e8b8d35b9658d8652a7857f69e30eb8.

promgen/serializers.py
vincent-olivert-riera: Fixed in 9d3afcb2a548d5de45f7c6037fe7b97b03a7a52d.

promgen/serializers.py
```diff
@@ -120,3 +121,55 @@ class HostSerializer(serializers.ModelSerializer):
...
+                if not any(
+                    (
+                        other_matcher["name"] in ["service", "project"]
+                        and other_matcher["isEqual"] == True
+                    )
+                    for other_matcher in data["matchers"]
+                ):
+                    rule = models.Rule.objects.get(name=matcher["value"])
+                    if rule.content_type.model == "site":
+                        raise errors.SilenceError.GLOBALSILENCE.error()
```
vincent-olivert-riera: @hoangpn , I have been testing and re-thinking about this.

The current code breaks when we try to create a silence like this: `alertname=~"foo.*"`.
This is because the code in line 157 will try to search for a rule named `foo.*` in the data base, which will not exist and therefore a `promgen.models.Rule.DoesNotExist` exception will be raised.

I think we should go much simpler on this, and just reject the creation of any silence that does not include either a `service` or `project` matcher.

Something like this (untested):

```
# Every silence should include either a service or project matcher
if not any(matcher["name"] in ["service", "project"] for matcher in data["matchers"]):
    raise errors.SilenceError.NOSERVICEORPROJECTMATCHER.error()
```

That `NOSERVICEORPROJECTMATCHER` error does not exist. It would need to be defined.

vincent-olivert-riera (changes requested):
@hoangpn , I have tested your branch and done another review spin.
I made one request. Everything else looks good :+1:

promgen/static/js/promgen.vue.js
```diff
@@ -121,6 +137,14 @@ const app = Vue.createApp({
             let tgt = document.getElementById(target);
             tgt.setAttribute('list', dst + '.' + src);
         },
+        getActiveSilencesForService: function (serivce) {
+            return getActiveSilences(this.activeSilences, 'service', serivce);
```
vincent-olivert-riera: There is a typo in these two lines: `serivce` should be `service`.

promgen/static/js/promgen.vue.js
```diff
@@ -121,6 +137,14 @@ const app = Vue.createApp({
...
+        },
+        getActiveSilencesForProject: function (project, service) {
+            const silencesMatchProject = getActiveSilences(this.activeSilences, 'project', project);
+            const silencesMatchService = getActiveSilences(this.activeSilences, 'service', service);
+            return silencesMatchProject.filter(silence => silencesMatchService.includes(silence));
```
vincent-olivert-riera: There is a problem here.

If I create a silence with this matcher only: `project="foo.*"`
When I visit the project detail page of any project starting with the string `foo`, it should display the "Active silences" button. However, that's not the case.

It works if I create another silence that has both `project` and `service` matchers.

Could you please investigate that behavior?

promgen/serializers.py
vincent-olivert-riera: Fixed in e8267af96cb9b30d1ee34d32d5aac0e722407e52

promgen/static/js/promgen.vue.js
vincent-olivert-riera: Fixed in 8741b2b6ca3ada29eb91281ff62991dec1681590

hoangpn:
@vincent-olivert-riera 
Thank you for your hard work! I've rebased all the fixup commits.
Now the PR is ready for the final review. 🙇

promgen/static/js/promgen.vue.js
vincent-olivert-riera: Fixed in https://github.com/line/promgen/commit/8741b2b6ca3ada29eb91281ff62991dec1681590

vincent-olivert-riera (approved):
LGTM

PR #624, opened 2025-07-24
Using only one UI for showing Active Silences

hoangpn:
Previously, we displayed the Active Silence list of a Service using a modal dialog. Now, the way we display the Active Silence list of a Project will also be done similarly, changing from displaying as a collapsible table to a modal dialog to unify the user experience.


vincent-olivert-riera (approved):
LGTM

PR #629, opened 2025-08-07
Activate a tab based on the hash value of the URL

hoangpn:
We have added a JS function that allows users to directly access a tab on the UI from a URL if the URL contains a hash with the value being the id of that tab.

### DIFF
--- a/promgen/static/js/promgen.js
+++ b/promgen/static/js/promgen.js
@@ -144,7 +144,12 @@ $(document).ready(function() {
         // Ideally we would directly update things that need to be updated
         // but a page redraw is a bit easier since that also allows us to
         // update our page messages
-        window.location = data.redirect
+        window.location = data.redirect;
+        if (data.redirect.includes('#')) {
+          // If the redirect contains a hash, the browser may not reload the page
+          // so we need to force a reload. Otherwise this code won't be reached.
+          window.location.reload();
+        }
       });
     } else {
       // If we click the cancel button, then we restore the old state and
--- a/promgen/static/js/promgen.js
+++ b/promgen/static/js/promgen.js
@@ -108,6 +108,21 @@ document.addEventListener("DOMContentLoaded", function () {
   });
 });
 
+// Activate a tab based on the hash value of the URL
+// For example, if the URL is `http://example.com#tab2`, it will activate the tab with id `tab2`.
+function activateTabFromHash() {
+  const hash = window.location.hash;
+  if (hash && $(hash).length) {
+    // Remove active class from all tabs and tab content
+    $(".nav-tabs li").removeClass("active");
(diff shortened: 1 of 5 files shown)


promgen/static/js/promgen.js
```diff
@@ -138,3 +138,22 @@ $(document).ready(function() {
...
+      $(".tab-pane").removeClass("active");
+
+      // Add active class to tab and tab content that matches the hash
+      $('a[href="' + hash + '"]').parent().addClass("active");
+      $(hash).addClass("active");
+    }
+  }
+
+  activateTabFromHash();
+});
```
vincent-olivert-riera, line 173: On line number `111`, there already is a `$(document).ready` function.
Please, define your `activateTabFromHash` function outside, and call it from inside.

promgen/static/js/promgen.js
```diff
@@ -138,3 +138,22 @@ $(document).ready(function() {
     }
   });
 });
+
+// Activate a tab based on the hash value of the URL
+// For example, if the URL is `http://example.com#tab2`, it will activate the tab with id `tab2`.
+$(document).ready(function () {
+  function activateTabFromHash() {
+    var hash = window.location.hash;
```
vincent-olivert-riera: `hash` is not reassigned, therefore it can be `const`.

vincent-olivert-riera (changes requested):
Review done.

promgen/static/js/promgen.js
```diff
@@ -129,12 +144,32 @@ $(document).ready(function() {
...
         // update our page messages
-        window.location = data.redirect
+        if (data.redirect.includes('#')) {
+          // If the redirect contains a hash, the browser will not reload the page
+          // so we need to force a reload
+          window.location = data.redirect;
+          window.location.reload();
+        } else {
+          window.location = data.redirect;
+        }
```
vincent-olivert-riera, line 152: I think is safe to only put the `reload` inside the `if`.

There are 3 scenarios (4 actually, but they can be reduced to 3):
- The browser detects that `data.redirect` is different than the current URL: reload happens automatically (with hash or without hash, it doesn't matter).
- Two browser detects that `data.redirect` is the same as the current URL, and there is **no** hash: reload happens automatically.
- Two browser detects that `data.redirect` is the same as the current URL, and there is a hash: reload does not happen automatically.

So, in theory, we could update `window.location` unconditionally. And after that, check if `data.redirect` includes a hash. If so, `reload`.

```
        window.location = data.redirect;
        // If the redirect contains a hash, the browser may not reload the page
        // so we need to force a reload. Otherwise this code won't be reached.
        if (data.redirect.includes('#')) {
          window.location.reload();
        }
```

Could you please test that?

vincent-olivert-riera (dismissed):

PR #630, opened 2025-08-08
Fix the bug that overrides AUTHENTICATION_BACKENDS not work

hoangpn:
We fixed a bug that AUTHENTICATION_BACKENDS setting from PROMGEN was not replaced Django settings. Loading overrides from PROMGEN to replace Django settings should always be at the end of the settings.py file. Therefore, we changed the location of the AUTHENTICATION_BACKENDS setting.

### DIFF
--- a/promgen/settings.py
+++ b/promgen/settings.py
@@ -217,14 +217,13 @@ try:
 except ImportError:
     MIDDLEWARE.remove("debug_toolbar.middleware.DebugToolbarMiddleware")
 
-
-# Load overrides from PROMGEN to replace Django settings
-for k, v in PROMGEN.pop("django", {}).items():
-    globals()[k] = v
-
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 AUTHENTICATION_BACKENDS = (
     "django.contrib.auth.backends.ModelBackend",
     "guardian.backends.ObjectPermissionBackend",
 )
+
+# Load overrides from PROMGEN to replace Django settings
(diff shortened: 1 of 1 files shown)


promgen/settings.py
```diff
@@ -217,14 +217,13 @@
...
+AUTHENTICATION_BACKENDS = (
+    "django.contrib.auth.backends.ModelBackend",
+    "guardian.backends.ObjectPermissionBackend",
+)
 
 # Load overrides from PROMGEN to replace Django settings
 for k, v in PROMGEN.pop("django", {}).items():
     globals()[k] = v
 
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
vincent-olivert-riera: We still have this `DEFAULT_AUTO_FIELD` variable **after** the overrides, so we won't be able to override it.

promgen/settings.py
```diff
@@ -217,14 +217,14 @@
...
-for k, v in PROMGEN.pop("django", {}).items():
-    globals()[k] = v
-
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 AUTHENTICATION_BACKENDS = (
     "django.contrib.auth.backends.ModelBackend",
     "guardian.backends.ObjectPermissionBackend",
 )
+
```
vincent-olivert-riera, line 226: Please remove this empty line.

After that, squash all commits into one.

vincent-olivert-riera (changes requested):
Review done.

PR #632, opened 2025-08-12
Feature/improve audit log

hoangpn:


promgen/models.py
```diff
@@ -599,6 +607,15 @@ def highlight(self):
             return "danger"
         return ""
 
+    @staticmethod
+    def find_parent_object(obj):
```
vincent-olivert-riera: I think specifying `_object` is redundant. And also, `get` is more appropriate than `find` in this case.

I would name this `get_parent`, as simple as that.

Later in the code, there are also a few more redundant things like:
- `instance_parent_object` -> `parent`
- `old_parent_object` -> `old_parent`

Please apply this patch (`git am <that file>`): 
[0001-fixup-Modify-the-Audit-model-to-improve-the-Audit-Lo.PATCH](https://github.com/user-attachments/files/21747712/0001-fixup-Modify-the-Audit-model-to-improve-the-Audit-Lo.PATCH)

promgen/migrations/0030_audit_parent_content_type_id_audit_parent_object_id_and_more.py
```diff
@@ -0,0 +1,29 @@
...
+            field=models.PositiveIntegerField(default=0),
+        ),
+        migrations.AddIndex(
+            model_name="audit",
+            index=models.Index(
+                fields=["parent_content_type_id", "parent_object_id"],
+                name="promgen_aud_parent__4cbb33_idx",
+            ),
+        ),
+    ]
```
vincent-olivert-riera, line 60: We should be able to create the new columns and populate them in the same migration file.

Please apply this patch: 
[0002-fixup-Modify-the-Audit-model-to-improve-the-Audit-Lo.PATCH](https://github.com/user-attachments/files/21747719/0002-fixup-Modify-the-Audit-model-to-improve-the-Audit-Lo.PATCH)

promgen/migrations/0031_migrate_parent_object_for_audit_log.py
```diff
@@ -0,0 +1,39 @@
...
+                    log.save()
+                    break
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("promgen", "0030_audit_parent_content_type_id_audit_parent_object_id_and_more"),
+    ]
+
+    operations = [migrations.RunPython(extract_data_and_populate_parent_object)]
```
vincent-olivert-riera: If we populate the columns in the other migration file, this one becomes unnecessary.
Please remove it: 
[0003-Revert-Add-a-migration-script-for-migrate-old-audit-.PATCH](https://github.com/user-attachments/files/21747736/0003-Revert-Add-a-migration-script-for-migrate-old-audit-.PATCH)

PR #634, opened 2025-08-14
Support change member's role on UI

hoangpn:
Currently, the only way to change a member's role is to delete the member and add them again with the desired role. From a UX perspective, this is somewhat inconvenient. Therefore, we have added a small button next to each member's role label in the member list table to quickly support changing the member's role.
And since users cannot remove the owner's permission without transferring
ownership, we also have removed the button for this action from the UI.

### DIFF
--- a/promgen/templates/promgen/permission_row.html
+++ b/promgen/templates/promgen/permission_row.html
@@ -22,6 +22,7 @@
     {% endfor %}
   </td>
   <td class="col-xs-3" style="word-break: break-all" v-pre>
+    {% if user.username != object.owner.username %}
     <form
       method="post"
       action="{% url 'permission-delete' %}"
@@ -48,5 +49,6 @@
       </div>
       {% endif %}
     </form>
+    {% endif %}
   </td>
 </tr>
--- a/promgen/templates/promgen/permission_row.html
+++ b/promgen/templates/promgen/permission_row.html
@@ -20,6 +20,23 @@
       {{ perm|upper }}
     </span>
     {% endfor %}
+
+    {% if user.username != object.owner.username %}
+    <button style="background:none" class="btn btn-xs" data-toggle="collapse" data-target="#user{{user.id}}">
+      <span class="glyphicon glyphicon-cog" aria-hidden="true"></span>
+    </button>
+    <div class="collapse" id="user{{user.id}}">
+      <form method="post" action="{% url 'permission-assign' %}" style="display: inline">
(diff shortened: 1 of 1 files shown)


promgen/templates/promgen/permission_row.html
```diff
@@ -20,6 +20,21 @@
       {{ perm|upper }}
     </span>
     {% endfor %}
+
+    <button class="btn btn-xs" data-toggle="collapse" data-target="#user{{user.id}}">
```
vincent-olivert-riera: I think it will look better if this button does not have a background. Can you try it?

```suggestion
    <button style="background:none" class="btn btn-xs" data-toggle="collapse" data-target="#user{{user.id}}">
```

promgen/templates/promgen/permission_row.html
```diff
@@ -20,6 +20,21 @@
...
+      <span class="glyphicon glyphicon-refresh" aria-hidden="true"></span>
```
vincent-olivert-riera: I think the refresh icon is not very appropriate. Since we are going to change a setting (in this case, the role of a member), I think it would be better if we use the cog icon. Please try it:

```suggestion
      <span class="glyphicon glyphicon-cog" aria-hidden="true"></span>
```

vincent-olivert-riera (changes requested):

promgen/templates/promgen/permission_row.html
```diff
@@ -20,6 +20,21 @@
...
+      <form method="post" action="{% url 'permission-assign' %}" style="display: inline">
+        {% csrf_token %}
+        <input name="next" type="hidden" value="{{ request.get_full_path }}#members" />
+        <input name="model" type="hidden" value="{{ object | klass }}" />
+        <input name="id" type="hidden" value="{{ object.id }}" />
+        <input name="username" type="hidden" value="{{ user.username }}" />
+        {{ permission_form.permission }}
+        <button class="btn btn-primary btn-xs">Change</button>
+      </form>
+    </div>
```
vincent-olivert-riera, line 38: @hoangpn , one more thing. The owner of an object should always have admin permissions, and you cannot change its permissions unless you change the object's ownership.
Therefore, this cog icon and the whole collapsed form should not be included if the user is the owner of the object.

```
{% if not owner %}
Display all of that
{% endif %}
```

vincent-olivert-riera (approved):

PR #639, opened 2025-08-22
Autocomplete select box for choosing user

hoangpn:
Currently, selecting a username or owner in the select box within forms is very inconvenient when there are hundreds of users in the system. We have used the JQuery Select2 to add autocomplete functionality to these selects. This will provide a better user experience.


promgen/templates/base.html
```diff
@@ -36,6 +37,7 @@
   <script src="{% static 'js/luxon.min.js' %}"></script>
   <script src="{% static 'js/linkify.min.js' %}"></script>
   <script src="{% static 'js/linkify-string.min.js' %}"></script>
+  <script src="{% static 'js/select2.min.js' %}?v=1"></script>
```
vincent-olivert-riera: The `?v=1` is for Javascript files that are not versioned, like our `promgen.js` file for instance.

But `select2` does release versions, so we should import it with the right name (including the version number): `js/select2-4.1.0-rc.0.min.js`.

Same for its CSS file.

Of course you will need to rename the real files under the `js` and `css` directories.

vincent-olivert-riera (changes requested):

promgen/templates/base.html
hoangpn: Thank you. I got it. 🙇

PR #641, opened 2025-09-04
Add the Group model

hoangpn:
Add the Group model, which is a collection of multiple users. Each user will have one of two roles: Admin or Member.
User can create group and manage member on Promgen Web.

image


promgen/templates/promgen/group_detail.html
```diff
@@ -0,0 +1,39 @@
...
+{% load promgen %}
+{% block title %}
+Promgen / Group / {{ group.name }}
+{% endblock %}
+
+{% block content %}
+
+<div class="page-header promgen-flex-space-between-center">
+  <div>
+    <h1>Group: {{ group.name }}</h1>
```
vincent-olivert-riera, line 12: User's can create groups with malicious names that execute Javascript code. We should protect this using `v-pre`, like we do in other places.

promgen/templates/promgen/group_detail.html
```diff
@@ -0,0 +1,39 @@
...
+  </div>
+  <div>
+    <a
+      href="{% urlqs 'audit-list' group=group.id %}"
+      class="btn btn-primary btn-sm"
+      >{% translate "View Edit History" %}</a
+    >
+    <a href="{% url 'group-update' group.id %}" class="btn btn-warning btn-sm"
+      >{% translate "Edit Group" %}</a
+    >
```
vincent-olivert-riera: The formatting of these couple of `a` elements went wrong. It should be like this:

```
    <a
      href="{% urlqs 'audit-list' group=group.id %}"
      class="btn btn-primary btn-sm"
    >
      {% translate "View Edit History" %}
    </a>
    <a href="{% url 'group-update' group.id %}" class="btn btn-warning btn-sm">
      {% translate "Edit Group" %}
    </a>
```

promgen/urls.py
```diff
@@ -90,6 +90,14 @@
...
     path("permission/assign", views.PermissionAssign.as_view(), name="permission-assign"),
     path("permission/delete", views.PermissionDelete.as_view(), name="permission-delete"),
+    # Groups
+    path("group", views.GroupList.as_view(), name="group-list"),
+    path("group/new", views.GroupRegister.as_view(), name="group-register"),
+    path("group/<int:pk>", views.GroupDetail.as_view(), name="group-detail"),
+    path("group/<int:pk>/add", views.GroupAddMember.as_view(), name="group-add"),
+    path("group/<int:pk>/remove", views.GroupRemoveMember.as_view(), name="group-remove"),
+    path("group/<int:pk>/update", views.GroupUpdate.as_view(), name="group-update"),
+    path("group/<int:pk>/delete", views.GroupDelete.as_view(), name="group-delete"),
```
vincent-olivert-riera, line 101: I think the URLs for the membership management should be more explicit (yeah, I know the other rules that already exist are not a good example of explicitness). But as they are now, `group/<int:pk>/remove` and `group/<int:pk>/delete` look like they both would delete the group itself.

How about this instead? :arrow_down: 

```
    path("group/<int:pk>/add-member", views.GroupAddMember.as_view(), name="group-add-member"),
    path("group/<int:pk>/remove-member", views.GroupRemoveMember.as_view(), name="group-remove-member"),
```

If the path with hyphen (`group/<int:pk>/add-member`) gives you problems, try with underscore (`group/<int:pk>/add_member`). But I think it should be OK.

promgen/views.py
```diff
@@ -1615,3 +1615,92 @@ def get_object(self):
...
+
+
+class GroupRemoveMember(LoginRequiredMixin, SingleObjectMixin, View):
+    model = models.Group
+
+    def post(self, request, *args, **kwargs):
+        group = self.get_object()
+        user = User.objects.get_by_natural_key(request.POST["username"])
+
+        group.user_set.remove(user)
```
vincent-olivert-riera: You could make a malicious POST request with a user that is not a member of the group.
So, I think before trying to remove a user from a group, we should first check that is actually a member of it. If it's not, then we should display a `messages.error` saying that the user is not a member of that group.

Also, when removing a user from a group, even if the current code works, it feels strange: first it removes the user from the group, and after that it removes its permissions for the group.
Wouldn't be more logical to do it in the reverse order? First remove all its permissions for that group, and finally, remove the user from the group.

promgen/views.py
```diff
@@ -1615,3 +1615,92 @@ def get_object(self):
...
+    def post(self, request, *args, **kwargs):
+        group = self.get_object()
+        users = User.objects.filter(username__in=request.POST.getlist("users"))
+        for user in users:
+            group.user_set.add(user)
+
+            permissions = get_perms(user, group)
+            for perm in permissions:
+                remove_perm(perm, user, group)
+            assign_perm(request.POST["permission"], user, group)
```
vincent-olivert-riera, line 1683: Uhm..., I'm a little bit unsure about this. The current code allows you to specify users that are already members of the group, and replace its permissions with the ones provided in the request.

I think for big groups this could cause issues where someone accidentally modifies the permissions of a user that was already a member.

If you want to change the permissions of a member, I think you should do it using the cog button.

If the request provides some users that are already members, we should detect that and don't try to add them or modify their permissions.
At the end, along with the `success` message for the newly added users, we could also display a `warning` message saying that users A, B and C were already members of the group. We should create two separate users list for those messages, one with the users that were already members, and another one with the users that were added.

promgen/migrations/0031_group.py
```diff
@@ -0,0 +1,48 @@
...
+
+def reverse_code(apps, schema_editor):
+    # Write reverse migration code if necessary
+    pass
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('auth', '0012_alter_user_first_name_max_length'),
+        ('promgen', '0030_audit_parent_content_type_id_audit_parent_object_id_and_more'),
```
vincent-olivert-riera, line 27: We use the `dependencies` to apply the migrations always in the same order. So, the only thing you need here is a dependency on the latest migration file (`0030`), nothing else. The dependency on `0012` can be removed.

vincent-olivert-riera (changes requested):
@hoangpn , review done.

promgen/migrations/0031_group.py
hoangpn: This migration file was auto-generated by Django. (line 26-28)  👀  Should we touch them?

promgen/views.py
hoangpn: Nice suggestion! I got it. 🙇

promgen/migrations/0031_group.py
vincent-olivert-riera: Yeah, remove the unnecessary dependency please :bow:

promgen/templates/promgen/group_detail.html
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/b861356a199fc4d0512dc05a303ccd9870a09ab7

promgen/templates/promgen/group_detail.html
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/ce8012afa104d5c4a4b198a2c00af7f3651fd67d

promgen/migrations/0031_group.py
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/f8009b137d28259d64b7d63d5f538363854e5591

promgen/urls.py
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/3b9802ec71d0bb1c92de7018dddf11392a670b9f

promgen/views.py
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/9bb2689c7c8dd65f9ecb8f400ca0575b188e0af7

promgen/views.py
hoangpn: Fixed in https://github.com/line/promgen/pull/641/commits/aa12668db209729a329ed5d6ecfd0c36f51664fe

promgen/views.py
```diff
@@ -1636,22 +1636,40 @@ class GroupAddMember(LoginRequiredMixin, SingleObjectMixin, View):
...
-                remove_perm(perm, user, group)
-            assign_perm(request.POST["permission"], user, group)
-        messages.success(
-            request,
-            _("Added {users} to {group} with permission {permission}").format(
-                users=[user.username for user in users],
-                group=group.name,
-                permission=request.POST["permission"],
-            ),
+        existed_members = User.objects.filter(
```
vincent-olivert-riera: Could you call this variable `already_members` instead?

vincent-olivert-riera (changes requested):
Thanks for all the fixups. It now looks great.
I have only one last (very minor) request; please have a look.

promgen/views.py
```diff
@@ -1673,6 +1669,35 @@ def post(self, request, *args, **kwargs):
...
+                remove_perm(perm, user, group)
+            assign_perm(request.POST["permission"], user, group)
+            messages.success(
+                request,
+                _("Updated permission to {permission} for {user} on Group {group}").format(
+                    permission=request.POST["permission"], user=user.username, group=group.name
+                ),
+            )
+        else:
+            messages.warning(
```
vincent-olivert-riera: Make this a `messages.error`. This should never happen, so, if it does, something is really wrong, or we have received a malicious request.

vincent-olivert-riera (approved):

PR #644, opened 2025-09-08
Integrating per-object-permission with the Group model

hoangpn:
We have updated the Members tab on Service Detail, Project Detail, and Farm Detail pages. Users can now assign or remove permissions for groups on specific Services, Projects, or Farms. Additionally, users can also view the list of Groups with current permissions on these objects.


promgen/static/js/promgen.js
```diff
@@ -109,9 +109,12 @@ document.addEventListener("DOMContentLoaded", function () {
 });
 
 function initSelect2() {
-  const selectors = 'select[name="username"],select[name="owner"]';
+  const selectors = ['username', 'owner', 'group']
```
vincent-olivert-riera:
```suggestion
  const selectors = ["username", "owner", "group"]
```

promgen/static/js/promgen.js
```diff
@@ -109,9 +109,12 @@ document.addEventListener("DOMContentLoaded", function () {
...
+    .map(name => `select[name="${name}"]`)
+    .join(',');
```
vincent-olivert-riera:
```suggestion
    .join(",");
```

promgen/static/js/promgen.js
```diff
@@ -188,6 +191,14 @@ $(document).ready(function() {
       activateTabFromHash();
   });
 
+  $('input[type="radio"][name="perm-type"]').change(function () {
+    var target = $(this).data("target");
```
vincent-olivert-riera:
```suggestion
    const target = $(this).data("target");
```

promgen/views.py
```diff
@@ -1570,53 +1592,75 @@ def get_object(self):
...
         models = ContentType.objects.get(app_label="promgen", model=model)
         obj = models.get_object_for_this_type(pk=id)
         return obj
 
+    def build_success_message(self, message):
+        if "on" == self.request.POST.get("remove_sub_permissions") and isinstance(
+            self.get_object(), models.Service
+        ):
+            return _(message + " and its projects.")
+        return _(message + ".")
```
vincent-olivert-riera, line 1662: There should be a few messages that need translation. Please, also include in this PR the changes to the `.po` and `.mo` files.

promgen/templates/promgen/permission_group_row.html
```diff
@@ -0,0 +1,91 @@
...
+          for="removeSubPermissions"
+          style="display: inline"
+        >
+          <small>{% trans "Remove permissions in child projects" %}</small>
+        </label>
+      </div>
+      {% endif %}
+    </form>
+  </td>
+</tr>
```
vincent-olivert-riera, line 85: image

1: The name of the group could be a link to the group detail page.
2: We don't need that column. If you want to see the members of the group, go to the group detail page. Groups with a large quantity of members could make that page look very ugly.
3: This bug was not introduced in this PR, but since the screenshot resulted that way, I will talk about it. As you can observe, the `child` word has been broken and continues in the second line. We should prevent that to happen, and break only on complete words. I think the `overflow-wrap: break-word` CSS should fix it, but please test it.

hoangpn: > The name of the group could be a link to the group detail page.

I was thinking about it when I created this table. Currently, on Promgen's tables, sometimes the name is linked to the object, while sometimes it uses the "Edit" button on the Actions column.
Which way do you think the best? Or both of them? 👀

hoangpn: > 3: This bug was not introduced in this PR...

I fixed by removing the 'word-break' style of the parent element @@

hoangpn: Fixed in https://github.com/line/promgen/pull/644/commits/066823f7e9bd4448af57c5464f984454b2f80d7c

promgen/views.py
hoangpn: I got it. I will add the necessary translations when the UI of this PR is ready. 🙇

promgen/templates/promgen/permission_group_row.html
vincent-olivert-riera: > > The name of the group could be a link to the group detail page.
> 
> I was thinking about it when I created this table. Currently, on Promgen's tables, sometimes the name is linked to the object, while sometimes it uses the "Edit" button on the Actions column. Which way do you think the best? Or both of them? 👀

I think we should have the same actions that we have for the users: only the "Delete" button.

vincent-olivert-riera: Fixed in 60039d14acbf9db8ff16d98da55c08b6066a3748

promgen/views.py
vincent-olivert-riera: Fixed in 832bb6af51cd55ca4f0e327eb6401fc11ad8a574

PR #648, opened 2025-09-12
Feature/restructure farm

hoangpn:
Convert the relationship between the **Project** and the **Farm** from **N-1** to **1-1** by reusing the farm_id field in the table promgen_project and **cloning** data in the table promgen_farm. 

This PR includes the following changes:

- Migration script for cloning farms.
- Remove the Farm Link feature for Local Farm.
- Remove the Farm's permission management feature.
- Remove the owner field of the Farm.
- Updates on UI due to BE's changes.

UI changes:

**Project's Hosts** : Buttons and Confirn message (only changes for Local Farm)
AS-IS:
image
image
image
TO-BE:
image
image
image

**Farm Register**: Remove _owner_ field
AS-IS:
image
TO-BE:
image

**Farm Detail**: Remove panels and change the breadcrumb
AS-IS:
image
TO-BE:
image

**Host Detail**:  Change UI and the Silence button
AS-IS:
image
TO-BE:
image

**Farm List**: Remove _ul_ tag due to only one project per farm
AS-IS:
image
TO-BE:
image

**Host List**: Remove _ul_ tag due to only one project per farm
AS-IS:
image
TO-BE:
image

**Search**:  Change UI and the Silence button
AS-IS:
image
TO-BE:
image


promgen/templates/promgen/search.html
```diff
@@ -119,15 +123,38 @@ <h1>Search</h1>
...
       <tr>
         <td><a href="{% url 'host-detail' host.name %}" v-pre>{{ host.name }}</a></td>
-        <td><a
-          @click="setSilenceDataset"
-          class="btn btn-warning btn-xs"
-          data-instance="{{host.name}}:[0-9]*"
-          style="display: inline-block;"
-        >{% trans "Silence" %}</a></td>
+        <td><a href="{% url 'farm-detail' host.farm.id %}" v-pre>{{ host.farm.name }}</a></td>
+        {% if host.farm.project_set.first %}
```
vincent-olivert-riera, line 136: At some point we will need to revisit this `farm.project_set`.
That made sense because a farm could be linked to multiple projects. But now, a farm is only linked to a single project, so the set will always contain one element (we don't need a set anymore).
But for this PR it's OK to leave it as it is.

promgen/views.py
```diff
@@ -626,6 +615,10 @@ def post(self, request, pk):
 
 class FarmLink(LoginRequiredMixin, View):
     def get(self, request, pk, source):
+        if source == discovery.FARM_DEFAULT:
+            messages.error(request, "Cannot link to local farm")
+            return HttpResponseRedirect(reverse("project-detail", args=[pk]) + "#hosts")
+
```
vincent-olivert-riera, line 620: In theory this should never happen, but is good to have this check in case someone tries to hack the system :+1:

vincent-olivert-riera (approved):
LGTM

PR #649, opened 2025-09-23
Disable signal when running the remove_farm_perms migration

hoangpn:
Promgen has a signal that creates Audit Logs every time a user removes the permission of another user. This is not necessary in the case where we run a migration script to clean up permission objects because they are automatically deleted by Promgen itself. This signal can even cause some unexpected errors if it is triggered in this case. Therefore, we temporarily disable this signal when running the migration script.


vincent-olivert-riera:
> @hoangpn ,
> 
> > This signal can even cause some unexpected errors if it is triggered in this case.
> 
> Please provide more details about that. Under which circumstances it happens, and which error message do you see?

After our private discussion between me and @hoangpn , we have decided that we will go with this commit message:

```
Disconnect signal while running the remove_farm_perms migration

At this moment, the real relationship between a Project and a Farm is one-to-one. To honor that, in
a future PR we plan to define the relationship between Project and Farm as a OneToOneField [1]. When
that happens, the project_id field on the Farm model will not exist anymore.

If the remove_farm_perms migration runs with that new code in place, the delete_log signal that is
triggered when the permissions are removed from a Farm instance will fail because at some point (in
order to construct a log) will try to access the project_id field of that Farm instance.

Taking into account that we do not care about Farm permissions anymore (because they will be
inherited from the Project to which it belongs), we also do not care about those logs.

Therefore, to allow that migration to run without any issues in the future, we are disconnecting the
offending signal while it runs.

1: https://docs.djangoproject.com/en/5.2/topics/db/examples/one_to_one/#one-to-one-relationships
```

vincent-olivert-riera (approved):

PR #652, opened 2025-09-26
[WIP] One-To-One relationship between Farm and Project

hoangpn:


promgen/migrations/0037_remove_project_farm_alter_farm_project.py
```diff
@@ -18,7 +18,7 @@ class Migration(migrations.Migration):
             model_name="farm",
             name="project",
             field=models.OneToOneField(
-                null=True, on_delete=django.db.models.deletion.CASCADE, to="promgen.project"
+                on_delete=django.db.models.deletion.CASCADE, to="promgen.project"
```
vincent-olivert-riera, line 21: The commit 3213819f2274c1955e450877dcd5e257f3fe2902 ("Change the way accesses a Project from a Farm") is removing `null=True` from this migration file, which was created by commit bd5e0005e3c0df0afa1cc4080921ab75dd60ecc8 ("Remove "farm" field from the Project model").

If this change was possible in commit bd5e0005e3c0df0afa1cc4080921ab75dd60ecc8, then is should be in commit bd5e0005e3c0df0afa1cc4080921ab75dd60ecc8.
Otherwise commit 3213819f2274c1955e450877dcd5e257f3fe2902 should create a new migration file to perform this change.

vincent-olivert-riera (changes requested):

promgen/migrations/0037_remove_project_farm_alter_farm_project.py
hoangpn: I got it. You're right. It should be in commit https://github.com/line/promgen/commit/bd5e0005e3c0df0afa1cc4080921ab75dd60ecc8

PR #723, opened 2026-04-06
Fix wrong classification of sending alert error

hoangpn:
At commit 359ca533, we added the categorize_error() function to classify the error_type of sending alert notification based on the HTTP code for the purpose of labeling metrics. However, there was a mistake causing error_type to always be "other_error". The reason is that the HTTP Requests library (requests) always returns False when executing the `bool (response)` statement in case of a response failure. (Ref: https://github.com/psf/requests/issues/2002)

We have fixed this issue and added the necessary tests for verification.


vincent-olivert-riera (approved):
LGTM

PR #729, opened 2026-05-19
Add metrics for monitoring requests sent to Promgen

hoangpn:
We have implemented new metrics to monitor the number and processing duration of requests sent to Promgen, categorized by endpoint, method, and status_code. These new metrics have been initialized using a migration script. The values of these metrics will be calculated and stored in the MySQL database after being processed in the middleware class.


vincent-olivert-riera (approved):
Looks good to me.

PR #733, opened 2026-05-25
Add new APIs for integrating with Notifier

hoangpn:
We have added several APIs to support users integrating with Promgen's Notifiers:
- List and filter Notifiers.
- Update/Partially update an existing Notifier.
- Add new filter to an existing Notifier.
- Delete filter of an existing Notifier.
- Delete an existing Notifier.

### DIFF
--- a/promgen/tests/notification/test_email.py
+++ b/promgen/tests/notification/test_email.py
@@ -20,6 +20,10 @@ class EmailTest(tests.PromgenTest):
         NotificationEmail.create(obj=one, value="foo@example.com", owner=self.user)
         NotificationEmail.create(obj=two, value="bar@example.com", owner=self.user)
 
+        # We disable all other senders to ensure that only our email sender is triggered
+        # during this test.
+        models.Sender.objects.exclude(sender=NotificationEmail.__module__).update(enabled=False)
+
         self.user = self.force_login(username="demo")
         permission = Permission.objects.get(codename="process_alert")
         self.user.user_permissions.add(permission)
--- a/promgen/filters.py
+++ b/promgen/filters.py
@@ -1,5 +1,10 @@
 import django_filters
+from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
+from django.db.models import F, Value
+from django.db.models.functions import Coalesce, NullIf
+
+from promgen import models
 
 
 class ShardFilter(django_filters.rest_framework.FilterSet):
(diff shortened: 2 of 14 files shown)


hoangpn:
I force-pushed to keep update with master. 🙇

promgen/tests/notification/test_email.py
```diff
@@ -15,15 +15,16 @@ def setUp(self):
...
+        models.Sender.objects.all().delete()
 
         NotificationEmail.create(obj=one, value="example@example.com", owner=self.user)
         NotificationEmail.create(obj=one, value="foo@example.com", owner=self.user)
         NotificationEmail.create(obj=two, value="bar@example.com", owner=self.user)
 
-        self.user = self.force_login(username="demo")
-        permission = Permission.objects.get(codename="process_alert")
-        self.user.user_permissions.add(permission)
-
```
vincent-olivert-riera: This kind of change is done across multiple test scripts.

Is this a fix/preparation independent of the addition of the new Notifier API?
If so, can we do it in a separate commit?

vincent-olivert-riera (dismissed):

promgen/tests/notification/test_email.py
hoangpn: > Is this a fix/preparation independent of the addition of the new Notifier API?

I added new Notifiers objects to `/fixtures/testcases.yaml` as test data for running testcases in `test_rest.py`.
However, there were existing tests that initialized Notifiers only for themselves. The expectations of these tests only based on their test data, too. 
So, currently I'm just clearing the data existed in `/fixtures/testcases.yaml` before running these old tests above to avoiding the effect. This is the simplest (and laziest) approach. 🙇 I choose not fixing the old tests because this is just a change of quantity, not a logical change.

Should I fix the old tests to match with the new initialized data? 
If not, should I still move these to a separate commit? 
Please let me know your opinion. Thank you!

vincent-olivert-riera: > Should I fix the old tests to match with the new initialized data?

Yes, please. You can do it on this PR as separate commits.
First fix whatever you need to fix, make all the necessary refactoring, etc.
And finally, add the Notifier API.

hoangpn: Thank you for your suggestions. 🙇 I have made the changes based on these suggestions. 
Please help me to re-review my PR.

vincent-olivert-riera (approved):
LGTM

PR #737, opened 2026-06-02
Add new APIs for integrating with Rule

hoangpn:
We have added several APIs to support users integrating with Promgen's Rules:
- List and filter Rules.
- Retrieve detail information of an existing Rule.
- Update/Partially update an existing Rule.
- Delete an existing Rule.


hoangpn:
I forced-pushed due to a redundant blank space. 🙇

vincent-olivert-riera (approved):
LGTM

PR #746, opened 2026-07-08
Update notifier API to return both value and alias fields

hoangpn:
Previously, the retrieving notifier API only responded with the "value" field. If a notifier had an "alias", the "value" field would return the alias, which could be ambiguous for users. To improve clarity, we have updated to always return both "value" and "alias" fields.  When an "alias" exists, the "value" field will be set to null.


vincent-olivert-riera (approved):
LGTM

PR #747, opened 2026-07-10
Add new APIs for integrating with URL

hoangpn:
We have added several APIs to support users integrating with Promgen's URL:
- List and filter URLs.
- List and filter URL Probes.
- Delete an existing URL.


vincent-olivert-riera (approved):
LGTM

PR #748, opened 2026-07-14
Fix bug that Promgen cannot serve Rapidoc JS file in Production

hoangpn:
In Production, Promgen uses a static file system storage backend that also saves hashed copies of the files it stores. During the creation of these hashed copies, Django tries to normalize the "sourceMappingURL" if it exists in any JS files. Unfortunately, Rapidoc's "sourceMappingURL" is already hashed, and there is no matching file in the same storage. Therefore, Django's converter ignores the "sourceMappingURL" in the hashed file, which causes the integrity hash to change.
To resolve this problem, we removed the "sourceMappingURL" and recalculated the integrity hash.
Ref: https://docs.djangoproject.com/en/4.2/_modules/django/contrib/staticfiles/storage/


promgen/templates/base.html
```diff
@@ -1,3 +1,4 @@
+{% load sri %}
```
vincent-olivert-riera: I think this is for a future PR, no?

hoangpn: Oh. Thank you so much! 🙇 
That's my mistake when I splited the commits.

vincent-olivert-riera (approved):
LGTM

PR #749, opened 2026-07-14
Add Django SRI

hoangpn:
We have integrated Django SRI into Promgen and replaced the "static" template tag with "sri_static" to serve all static files with Subresource Integrity in the Production environment.


vincent-olivert-riera (approved):
LGTM

PR #752, opened 2026-07-22
Add new APIs for integrating with Group

hoangpn:
We have added several APIs to support users integrating with Promgen's Groups:
- List and filter Group.
- Retrieve detailed information about an existing Group.
- Update or partially update an existing Group.
- Delete an existing Group.
- List members of an existing Group.
- List assigned resources of an existing Group.
- Add members to an existing Group.
- Change role of a member in an existing Group.
- Remove a member from an existing Group.


vincent-olivert-riera (approved):
LGTM

PR #755, opened 2026-07-23
Add new APIs for integrating with Project

hoangpn:
We have added several APIs to support users integrating with Promgen's Project:
- List and filter Project.
- Retrieve detailed information about an existing Project.
- Update or partially update an existing Project.
- Delete an existing Project.
- List users and groups that have permissions on an existing Project.
- Assign user or group to an existing Project.
- Remove user or group from an existing Project.
- Register Exporter/Farm/Notifier/Rule/URL to an existing Project.


vincent-olivert-riera (approved):
LGTM

PR #758, opened 2026-08-03
Add new APIs for integrating with Service

hoangpn:
We have added several APIs to support users integrating with Promgen's Service:
- List and filter Service.
- Register new Project to an existing Service.
- Retrieve detailed information about an existing Service.
- Update or partially update an existing Service.
- Delete an existing Service.
- List users and groups that have permissions on an existing Service.
- Assign user or group to an existing Service.
- Remove user or group from an existing Service.
- Register new Notifier to an existing Service.


vincent-olivert-riera (approved):
LGTM 👍

PR #760, opened 2026-08-05
Add new APIs for integrating with User

hoangpn:
We have added several APIs to support users integrating with Promgen's User:
- List and filter User.
- Retrieve detailed information about the current User.
- Register new Notifier to the current User.


vincent-olivert-riera (approved):
LGTM

PR #761, opened 2026-08-06
Add new APIs for integrating with Shard

hoangpn:
We have added several APIs to support users integrating with Promgen's Shard:
- List and filter Shard.
- Retrieve detailed information about an existing Shard.


vincent-olivert-riera (approved):
LGTM

PR #762, opened 2026-08-06
Add new APIs for integrating with Site

hoangpn:
We have added several APIs to support users integrating with Promgen's Site:
- Retrieve detailed information about the current Site.
- Register new Rule to the current Site.


vincent-olivert-riera (approved):
LGTM

PR #763, opened 2026-08-06
Change the serializer class of Registering Rule API

hoangpn:
We have changed to use RuleRetrieveDetailSerializer for the Registering new Rule API, instead of using RuleRetrieveSimpleSerializer class. There are two reasons for this change:
- RuleRetrieveDetailSerializer has the "parent" field as read_only_field. We should not allow user to set the parent Rule when registering a new Rule because it is only added when an existing Rule is overridden.
- RuleRetrieveDetailSerializer includes the fields "annotations," "description," and "labels." These fields are useful for users. Users should be able to set them when registering a new Rule.

### DIFF
--- a/promgen/rest_v2.py
+++ b/promgen/rest_v2.py
@@ -303,13 +303,13 @@ class RuleMixin:
     @extend_schema(
         summary="Register Rule",
         description="Register a new rule for the specified object.",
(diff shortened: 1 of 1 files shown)


vincent-olivert-riera (approved):
LGTM

PR #764, opened 2026-08-11
Add new APIs for overriding Rule

hoangpn:
- Replace Rule's copy_to() by override() function, providing greater flexibility in usage. As a result, multiple instances of "overwrite" term are replaced by "override".
- Create new APIs for overriding Rules for Service and Project based on the newly created function.


vincent-olivert-riera (approved):
LGTM 👍

PR #766, opened 2026-08-12
Specify field type for Rule serializer's labels and annotations

hoangpn:
We have specified the type for labels and annotations in the Rule serializer to be a dictionary of strings. This change provides more detailed descriptions for Rule APIs's requests and responses in the API schema, as well as improving generated examples of these APIs.

**AS-IS:**
image
image

**TO-BE:**
image
image


vincent-olivert-riera (approved):
LGTM

PR #768, opened 2026-08-17
Improve Actions of the Rule List table on the HostDetail page

hoangpn:
- Refactor Actions of the Rule List table on the HostDetail page: Removing the Override button and adding Toggle button.
- Fix UI quirks of the Rule List table on the HostDetail page

### DIFF
--- a/promgen/templates/promgen/host_detail.html
+++ b/promgen/templates/promgen/host_detail.html
@@ -112,7 +112,7 @@
     </tr>
     {% current_role service.grouper as grouper_role %}
     {% if grouper_role == "ADMIN" or grouper_role == "EDIT" %}
-      {% include "promgen/rule_block.html" with rule_list=service.list modify=1 csrf_token=csrf_token only %}
+      {% include "promgen/rule_block.html" with rule_list=service.list toggle=1 modify=1 csrf_token=csrf_token only %}
     {% else %}
     {% include "promgen/rule_block.html" with rule_list=service.list csrf_token=csrf_token only %}
     {% endif %}
--- a/promgen/templates/promgen/host_detail.html
+++ b/promgen/templates/promgen/host_detail.html
@@ -100,9 +100,8 @@
 
 <div class="panel panel-default">
   <div class="panel-heading">Rules</div>
-  <table class="table table-bordered table-condensed">
-    {% include "promgen/rule_header.html" %}
     {% for service in service_list %}
+  <table class="table table-bordered table-condensed">
     <tr v-pre>
       <td colspan="5">
         <h2>
@@ -111,13 +110,15 @@
       </td>
     </tr>
     {% current_role service.grouper as grouper_role %}
+    {% include "promgen/rule_header.html" with role=grouper_role %}
     {% if grouper_role == "ADMIN" or grouper_role == "EDIT" %}
       {% include "promgen/rule_block.html" with rule_list=service.list toggle=1 modify=1 csrf_token=csrf_token only %}
     {% else %}
     {% include "promgen/rule_block.html" with rule_list=service.list csrf_token=csrf_token only %}
     {% endif %}
-    {% endfor %}
+    <tr></tr>
   </table>
+    {% endfor %}
 </div>
 {% endif %}
 
--- a/promgen/templates/promgen/host_detail.html
+++ b/promgen/templates/promgen/host_detail.html
@@ -112,9 +112,9 @@
     </tr>
     {% current_role service.grouper as grouper_role %}
     {% if grouper_role == "ADMIN" or grouper_role == "EDIT" %}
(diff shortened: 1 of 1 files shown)


promgen/templates/promgen/host_detail.html
```diff
@@ -111,13 +110,15 @@ <h2>
       </td>
     </tr>
     {% current_role service.grouper as grouper_role %}
+    {% include "promgen/rule_header.html" with role=grouper_role %}
     {% if grouper_role == "ADMIN" or grouper_role == "EDIT" %}
-      {% include "promgen/rule_block.html" with rule_list=service.list modify=1 override_id=service.grouper.id csrf_token=csrf_token only %}
+      {% include "promgen/rule_block.html" with rule_list=service.list toggle=1 delete=1 modify=1 csrf_token=csrf_token only %}
```
vincent-olivert-riera: Exclusively talking about commit d6b7939f6c2058921bc7f573717d9f3806375474.

I understand why you are removing `override_id=service.grouper.id`. But, why are you adding `toggle=1 delete=1`?

hoangpn: @vincent-olivert-riera 
I want to show the same buttons as on the Service/Project Detail pages, so I copied the parameters from there. These parameters have already been existed in the past. 👀

vincent-olivert-riera: But commit d6b7939 is only for disabling the override feature on the host detail page, right? Those changes don't seem to belong to that commit.

Also, please change the commit title. This is why:

The commit title currently says "Fix bug that user cannot use the Override button [...]".
If the commit fixes the bug, then, after the commit the user can use the override button, right?
But the commit is not actually doing that. The commit is removing the override feature on the host detail page.

hoangpn: I understand. I will change the commit's title and description. 🙇

hoangpn: @vincent-olivert-riera 
As we discussed, I separated the commits, update the commit title and description, and also changed the PR's description.
Please continue reviewing my PR. Thank you for your hard work! 🙇

vincent-olivert-riera: @hoangpn , it looks like after your last changes the `delete=1` addition is no longer present. Was that intentional?

hoangpn: @vincent-olivert-riera 
I checked the template carefully and saw no use of the `delete` parameter. It might be existed in the past.
So, it's redundant now and I did not add it.

promgen/templates/promgen/host_detail.html
```diff
@@ -111,13 +110,15 @@ <h2>
...
+    {% include "promgen/rule_header.html" with role=grouper_role %}
     {% if grouper_role == "ADMIN" or grouper_role == "EDIT" %}
-      {% include "promgen/rule_block.html" with rule_list=service.list modify=1 override_id=service.grouper.id csrf_token=csrf_token only %}
+      {% include "promgen/rule_block.html" with rule_list=service.list toggle=1 modify=1 csrf_token=csrf_token only %}
     {% else %}
-    {% include "promgen/rule_block.html" with rule_list=service.list override_id=service.grouper.id csrf_token=csrf_token only %}
+    {% include "promgen/rule_block.html" with rule_list=service.list csrf_token=csrf_token only %}
     {% endif %}
-    {% endfor %}
+    <tr></tr>
```
vincent-olivert-riera, line 119: Why are we adding an empty row here?

promgen/templates/promgen/host_detail.html
vincent-olivert-riera: Excellent :+1: Thanks.

promgen/templates/promgen/host_detail.html
hoangpn: This is a trick. I got a strange error. The bottom border disappeared. 👀  I tried adding one more row and it worked like a charm. Do you have any other solution? 🙇 
image

vincent-olivert-riera: It's because this CSS matcher...

```
.panel>.table-bordered>tbody>tr:last-child>td
```

...has this style:

```
border-bottom: 0;
```

If we change that `0` to `1px` the problem is fixed, without having to add an empty row.

Give it a try for a few minutes. If it's too difficult or too messy, let me know and I will accept the "empty row" solution. Although in that case, it would be nice to have an HTML comment explaining why do we have that empty row.

vincent-olivert-riera: We can probably add something in `promgen.css` to override that style for the rule tables.

hoangpn:
I close the PR due to change of scope. Another PR will be created.

PR #773, opened 2026-08-25
Support for Logging in API V2

hoangpn:
We have added a feature that allows the backend of Promgen to log request and response information for API V2. This is intended to display what has been done through these APIs. A unique trace ID is generated to link the request and response logs.

For **requests**, the method, path, and request body are logged. However, fields in the request body are truncated to avoid filling up the log. By default, the character limit is set to 128, but fields like "clause" are still logged in full as they are necessary. For **responses**, only the status and response content size are logged.

By default, this feature is disabled. It can be enabled by setting ENABLE_V2_API_LOGGING to True.

### DIFF
--- a/promgen/settings.py
+++ b/promgen/settings.py
@@ -230,6 +230,9 @@ AUTHENTICATION_BACKENDS = (
 # Maximum time to wait for all scraping operations to complete (in seconds)
 PROMGEN_EXPORTER_SCRAPE_TIMEOUT = env.int("PROMGEN_EXPORTER_SCRAPE_TIMEOUT", default=25)
 
+# v2 API settings
+V2_API_LOGGING_ENABLED = env.bool("V2_API_LOGGING_ENABLED", default=True)
+
 # Load overrides from PROMGEN to replace Django settings
 for k, v in PROMGEN.pop("django", {}).items():
     globals()[k] = v
--- a/promgen/middleware.py
+++ b/promgen/middleware.py
@@ -17,8 +17,10 @@ files, we need to handle some deduplication. This is handled by using the django
 caching system to set a key and then triggering the actual event from middleware
 """
 
+import json
 import logging
 import socket
+import uuid
 from datetime import datetime
 from threading import local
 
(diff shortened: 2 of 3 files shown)


promgen/settings.py
```diff
@@ -230,6 +230,9 @@
 # Maximum time to wait for all scraping operations to complete (in seconds)
 PROMGEN_EXPORTER_SCRAPE_TIMEOUT = env.int("PROMGEN_EXPORTER_SCRAPE_TIMEOUT", default=25)
 
+# Enable logging for v2 API requests
+ENABLE_V2_API_LOGGING = env.bool("ENABLE_V2_API_LOGGING", default=False)
```
vincent-olivert-riera: API logging is quite important and I think it should be enabled by default.

We can simplify the comment removing redundant stuff such as "Enable logging" (the variable name is self-explanatory) and "requests" (it's obvious that the logging is for the API requests; what else could it be?).

Normally, configuration systems name the variables by identifying the component first, followed by the specific attribute or state. That way multiple variables for the same component can be nicely grouped together and alphabetically sorted for better reading. For instance:

```
V2_API_LOGGING_ENABLED = True
V2_API_MAX_CONCURRENCY = 10
V2_API_TIMEOUT_SECONDS = 60
```

```suggestion
# v2 API settings
V2_API_LOGGING_ENABLED = env.bool("ENABLE_V2_API_LOGGING", default=True)
```

vincent-olivert-riera (changes requested):

promgen/settings.py
hoangpn: I understand. It really looks clearer with these standards, and also will be useful for future expansion. 🙇  Thank you so much for your suggestion!

hoangpn: @vincent-olivert-riera 
About enabling by default, do you worry about flooding the logs if we do that?
By the way, currently, I'm setting the log level set to `info`. Do you think this is appropriate? 🙇

hoangpn:
@vincent-olivert-riera 
I have made the changes following by your recommendations in our discussion. 
Please feel free to review my pull request. Thank you! 🙇

vincent-olivert-riera (approved):
LGTM

promgen/settings.py
vincent-olivert-riera: We discussed all of this in Slack.

PR #774, opened 2026-08-27
Add Custom Exception Handler for Django REST Framework

hoangpn:
By default, Django REST Framework (DRF) exception handler only intercepts known API exception (like ValidationError or PermissionDenied). If an unexpected Python exception occurs, it returns None, which causes Django to re-raise the error and return unexpected response. Therefore, we have added a custom exception handler to safely catch unexpected exceptions and return a standard 500 Server Error with a structured JSON response.

### DIFF
--- a/promgen/tests/test_rest.py
+++ b/promgen/tests/test_rest.py
@@ -570,3 +570,30 @@ class RestAPITest(tests.PromgenTest):
         cases = tests.Data("cases", "test_rest_site.csv").csv()
         for case in cases:
             self._run_rest_test(case)
+
+    @override_settings(PROMGEN=tests.SETTINGS)
+    def test_exception_handler(self):
+        admin = User.objects.get(username="admin")
+        admin_token = Token.objects.filter(user=admin).first().key
+
+        # Expected exception (rest_framework.ValidationError) returns 400 HTTP code
+        # with the detail response.
+        response = self.client.patch(
+            reverse("api-v2:rule-detail", kwargs={"id": 1}),
+            data={"name": ""},
+            content_type="application/json",
+            HTTP_AUTHORIZATION=f"Token {admin_token}",
+        )
+        self.assertEqual(response.status_code, 400)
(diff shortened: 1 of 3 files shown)


promgen/tests/test_rest.py
```diff
@@ -570,3 +570,30 @@ def test_rest_site(self):
...
+        # Unexpected exception (django.ValidationError) returns 500 HTTP code with
+        # the default response.
+        response = self.client.patch(
+            reverse("api-v2:rule-detail", kwargs={"id": 1}),
+            data={"clause": "This clause is not valid."},
+            content_type="application/json",
+            HTTP_AUTHORIZATION=f"Token {admin_token}",
+        )
+        self.assertEqual(response.status_code, 500)
+        self.assertEqual(response.json(), {"error": "Server Error (500)"})
```
vincent-olivert-riera, line 599: I'm fine by merging this PR as it is, because it's an improvement on what we had before.

However, that error specifically should return this JSON instead: `{"error": "The clause is not valid."}` with a `400` status code.

Do you want to do it on this PR, or do you prefer to do it in the future?

vincent-olivert-riera (approved):

promgen/tests/test_rest.py
hoangpn: @vincent-olivert-riera 
Thank you for your review. I prefer to do it in the future. 🙇 
The reason is, this PR is only for handling unexpected exceptions. These can be various types of exceptions such as NullPointer, KeyError, database crash, etc.
In this PR, I want to use `django.ValidationError` as a test case so that everyone can understand that, although it is still a `ValidationError`, it is thrown by `Django` instead of `rest_framework`, so it remains an unexpected exception for `rest_framework`. Of course, I also note that this case would be more appropriate if we return a 400 and specify that the clause field is invalid. This will be done in another PR, and I will replace this test case with that one at a later time. Is this fine with you? 🙇 Please let me know your opinion.

vincent-olivert-riera: No problem at all. Let's merge this :)

PR #775, opened 2026-09-07
Feature: Knox Authentication

hoangpn:
Replacing the default authentication class in Django REST Framework (DRF) with Knox's TokenAuthentication instead of DRF's TokenAuthentication. This brings some new features to Promgen API Token Authentication.
- Allow multiple tokens per user.
- Configurable the number of tokens per user.
- Token expiration.
- Configurable token expiration in days.
- **[Security Improvement]** Not store tokens in database.

Existing DRF tokens will be migrate to Knox tokens with default expiration. User can continue using their created tokens without any issues.
- Tested migration script with empty database.
- Tested migration script with existing database + existing tokens.
- Tested migration script with existing database + no existing tokens.


promgen/settings.py
```diff
@@ -234,6 +234,15 @@
 # v2 API settings
 V2_API_LOGGING_ENABLED = env.bool("V2_API_LOGGING_ENABLED", default=True)
 
+# When set, Promgen will not allow the creation of API tokens with expiry greater than this setting.
+# None means unlimited.
+API_TOKEN_EXPIRATION_DAY_LIMIT = env.int("API_TOKEN_EXPIRATION_DAY_LIMIT", default=None)
```
vincent-olivert-riera:
```suggestion
API_TOKEN_EXPIRATION_LIMIT_DAYS = env.int("API_TOKEN_EXPIRATION_LIMIT_DAYS", default=None)
```

Placing the measurement unit (`_DAYS`, `_SECONDS`, `_MB`, `_BYTES`) at the very end of a variable name is a standard naming convention because it makes the scope clear to developers at a glance: `COMPONENT_PROPERTY_UNIT`.
In this case:
- `COMPONENT` = `API_TOKEN`
- `PROPERTY` = `EXPIRATION_LIMIT`
- `UNIT` = `DAYS`

`DAY_LIMIT` can easily be misread as a daily rate limit (e.g., how many requests or tokens can be used per day).
`LIMIT_DAYS` explicitly clarifies that the limit itself is expressed in a number of days.

I don't have a very strong preference for that variable name, as long as it follows that convention and its easy to understand. For instance, other options could be `API_TOKEN_TTL_DAYS` or `API_TOKEN_MAX_AGE_DAYS`.

Feel free to choose the one you like the most 👍

promgen/static/js/promgen.js
```diff
@@ -245,4 +245,30 @@ $(document).ready(function() {
...
+    navigator.clipboard.writeText(copyText).then(function() {
+      $("#copy-to-clipboard-btn")
+        .attr("data-original-title", "Copied!")
+        .tooltip('show');
+      // Reset the tooltip text after mouse leaves the button
+      $("#copy-to-clipboard-btn").mouseleave(function() {
+        $(this).attr("data-original-title", "Copy to clipboard");
+      });
+    })
+  });
```
vincent-olivert-riera, line 262: I don't want to delay this PR more, so keep this task for the future.

Can we make this a Vue component so it can be reused?

As it is right now, you need the ID of the button to be exactly `copy-to-clipboard-btn`, and the ID of the element you want to copy its value to be `copy-to-clipboard-input`. That means you can only have one of these buttons in the same page because in HTML the IDs must be unique.

If this was a Vue component we could put as many `<CopyToClipboard>` buttons as we want in the same page, and in others. For instance:
- `<CopyToClipboard target-id="input-one-id" />`
- `<CopyToClipboard target-id="input-two-id" />`

You could make it even more reusable by not only choosing the element that you want to copy from, but also which attribute or property you want to copy. For instance:
- `<CopyToClipboard target-id="input-one-id" target-attr="value">` (this copies the `value` attribute of an `input` element).
- `<CopyToClipboard target-id="label-one-id" target-prop="textContent">` (this copies the `textContent` property of a `label` element.
- Etc.

promgen/templates/promgen/profile.html
```diff
@@ -8,54 +8,57 @@ <h1>{{user.username}} ({{user.email}})</h1>
 </div>
 
 <div class="panel panel-primary" v-pre>
-  <div class="panel-heading">API Token</div>
+  <div class="panel-heading">API Tokens</div>
   <table class="table">
     <tr>
-      <th>Token</th>
+      <th>Name</th>
+      <th>Token Prefix</th>
```
vincent-olivert-riera: Uhm..., I would rather hide this information (the token prefix). As far as I can see, Grafana does not display any part of the actual token, so I think we don't need to do it either.

promgen/templates/promgen/profile.html
```diff
@@ -8,54 +8,57 @@ <h1>{{user.username}} ({{user.email}})</h1>
...
+      <th>Expires At</th>
```
vincent-olivert-riera, line 16: Can we also display the `Created At` column?

vincent-olivert-riera (changes requested):
@hoangpn , review done. Overall looks very good, but I have a few requests. Please have a look :)

promgen/static/js/promgen.js
hoangpn: I understand. Thank you so much for your suggestion!
Let's keep it as current, and I will create the Vue component in another PR. 🙇

promgen/settings.py
hoangpn: Thank you. I agree with your recommendation. 🙇 I will change it to `API_TOKEN_TTL_DAYS` since it has similarity with Knox's variables name.

promgen/templates/promgen/profile.html
hoangpn: Fine by me. I have referred to Grafana, GitHub, Atlassian and ChatGPT before designing our feature. Only ChatGPT shows the token prefix. On the other hand, Knox stores the token prefix although there will be no issues if it is not stored. Therefore, I guessed they encourage taking advantage of the stored token prefix.

If we don't display any part of the actual token, user will need to store their token name in somewhere in case they want to check the expiration of their token. It's not really a problem... so it's up to you. 😅

vincent-olivert-riera: > If we don't display any part of the actual token, user will need to store their token name in somewhere in case they want to check the expiration of their token. It's not really a problem... so it's up to you.

Yeah, the user is responsible for keeping track of their tokens. Let's hide the prefix.

promgen/templates/promgen/profile.html
hoangpn: Fixup commit: [c5498b0](https://github.com/line/promgen/pull/775/commits/c5498b0625cc66528b150f8570de714df8d82dde)

promgen/templates/promgen/profile.html
hoangpn: Fixup commit: [c5498b0](https://github.com/line/promgen/pull/775/commits/c5498b0625cc66528b150f8570de714df8d82dde)

promgen/settings.py
hoangpn: Fixup commits: 
[8741428](https://github.com/line/promgen/pull/775/commits/8741428d6dc156db043720304c96a45cbe16d05b)
[eda4848](https://github.com/line/promgen/pull/775/commits/eda4848f1cf0d7242dc923825e46f8d8a94141a5)

vincent-olivert-riera (approved):
LGTM

@hoangpn , please rebase and squash the fixups.

hoangpn:
@vincent-olivert-riera 
Thank you so much for your hardworking. The PR is ready for the final review. 🙇

PR #777, opened 2026-09-15
Feature: Custom Label

hoangpn:
Add a new feature allow Promgen Admin to create custom fields for Service and Project models. After that, they will be displayed in **Forms on UI** and **Components in API Schema**.
If user input value for these fields, they could be added to metrics when Prometheus scrapes them.


promgen/migrations/0051_customlabel_customlabelinstance.py
```diff
@@ -0,0 +1,99 @@
...
+                (
+                    "label_name",
+                    models.CharField(
+                        max_length=128,
+                        validators=[
+                            django.core.validators.RegexValidator(
+                                "^[a-zA-Z_][a-zA-Z0-9_]*$",
+                                "Only alphanumeric characters are allowed.",
+                            ),
+                            promgen.validators.custom_label_name,
```
vincent-olivert-riera, line 30: Can't we put all the rules inside this custom validator?

We want to stick with the Prometheus' [recommended](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) character set for label names, which is exactly that regex you just wrote above (`^[a-zA-Z_][a-zA-Z0-9_]*$`).
But, also, any label name starting with double underscore (`__`) is forbidden because is reserved for internal Prometheus use.
And also, we have our own set of labels that we want to forbid, like `service`, `project`, etc.

It would be nice to have all of those rules inside one single validator we can reuse everywhere.

hoangpn: I'm using existing custom validator in the `validators.py` file. The code above was auto-generated after changing `models.py` and running `makemigrations`.  👀 
You can see my original code here:
https://github.com/line/promgen/pull/777/changes/a61122d6d2156cab32b773b376dc910fe6e09c7f#diff-34ec5dcb4778d6385887806e45d80fb73abf7934cfde595d225182514abdbd15R807-R810

hoangpn:
@vincent-olivert-riera 
I added a fixup commit [2949bc1](https://github.com/line/promgen/pull/777/commits/2949bc18ddc22856128a1bba5e8561318d46ba21) to fix warning logs when running Promgen.

> promgen/serializers.py: Warning [ProjectViewSet > ProjectRetrieveDetailSerializer > ProjectCustomLabelSerializer]: Encountered 2 components with identical names "ProjectCustomLabel" and different identities <class 'promgen.serializers.ProjectCustomLabelSerializer'> and <class 'promgen.serializers.ProjectCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.
> promgen/serializers.py: Warning [ProjectViewSet > ProjectSimpleSerializer > ProjectCustomLabelSerializer]: Encountered 2 components with identical names "ProjectCustomLabel" and different identities <class 'promgen.serializers.ProjectCustomLabelSerializer'> and <class 'promgen.serializers.ProjectCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.
> promgen/serializers.py: Warning [ServiceViewSet > RegisterServiceSerializer > ServiceCustomLabelSerializer]: Encountered 2 components with identical names "ServiceCustomLabel" and different identities <class 'promgen.serializers.ServiceCustomLabelSerializer'> and <class 'promgen.serializers.ServiceCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.
> promgen/serializers.py: Warning [ServiceViewSet > ServiceRetrieveDetailSerializer > ServiceCustomLabelSerializer]: Encountered 2 components with identical names "ServiceCustomLabel" and different identities <class 'promgen.serializers.ServiceCustomLabelSerializer'> and <class 'promgen.serializers.ServiceCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.
> promgen/serializers.py: Warning [ServiceViewSet > ServiceRetrieveSimpleSerializer > ServiceCustomLabelSerializer]: Encountered 2 components with identical names "ServiceCustomLabel" and different identities <class 'promgen.serializers.ServiceCustomLabelSerializer'> and <class 'promgen.serializers.ServiceCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.
> promgen/serializers.py: Warning [ServiceViewSet > RegisterProjectToServiceSerializer > ProjectCustomLabelSerializer]: Encountered 2 components with identical names "ProjectCustomLabel" and different identities <class 'promgen.serializers.ProjectCustomLabelSerializer'> and <class 'promgen.serializers.ProjectCustomLabelSerializer'>. This will very likely result in an incorrect schema. Try renaming one.

The reason why we have these warnings is because I dynamically create the CustomLabelSerializer class to generate the correct API schema for corresponding models. However, some generated schemas may be identical, which is why we receive this warning. 

Therefore, I use an LRU cache to avoid duplication. (This solution came after discussing with AI.) Please feel free to review it. 🙇

promgen/migrations/0051_customlabel_customlabelinstance.py
vincent-olivert-riera: Yeah, I see way. It is because you are using two different validators for that model.
Ok, let me make another comment in the code.

promgen/validators.py
```diff
@@ -26,6 +26,33 @@
...
+        "name",
+        "owner",
+        "owner_id",
+        "pk",
+        "project",
+        "service",
+        "service_id",
+        "shard",
+        "shard_id",
+    ]
```
vincent-olivert-riera: Every time you call the `custom_label_name` function, you will define this list.
Better to extract the list outside the function so it's defined only once.

promgen/validators.py
```diff
@@ -26,6 +26,33 @@
...
+
```
vincent-olivert-riera, line 49: A custom label name must abide to the same rules as a normal label name, plus others that we define here.

Therefore, the first this you should do inside this validator is calling the other validator:

```
# Custom labels also follow the basic label rules
labelname(value)
```

promgen/validators.py
```diff
@@ -26,6 +26,33 @@
...
+    if value in RESERVED_LABEL_NAMES:
+        reserved_values = ", ".join(RESERVED_LABEL_NAMES)
+        raise ValidationError(f"Custom label name cannot be one of: {reserved_values}.")
+
+
```
vincent-olivert-riera, line 58: Also, validate the label name does not start with double underscore:

```
if value.startswith("__"):
    raise ValidationError(f"Custom label name cannot start with __")
```

promgen/models.py
```diff
@@ -801,3 +864,60 @@ def is_expired(self):
             if self.expiry < timezone.now():
                 return True
         return False
+
+
+class CustomLabel(models.Model):
+    label_name = models.CharField(
+        max_length=128,
+        validators=[validators.labelname, validators.custom_label_name],
```
vincent-olivert-riera: Only the `custom_label_name` validator should be enough, after all the modifications from my other comments.

vincent-olivert-riera (changes requested):
I just did a small review spin. I will do another one after deeper testing.

promgen/migrations/0051_customlabel_customlabelinstance.py
```diff
@@ -0,0 +1,95 @@
...
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models
+
+import promgen.validators
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("contenttypes", "0002_remove_content_type_name"),
```
vincent-olivert-riera: Why do (almost) all the migration files you are adding depend on `0002_remove_content_type_name`?
Migration files should be applied sequentially, so they just need to depend on the previous one.
Could you please fix that?

vincent-olivert-riera (changes requested):
Just one comment about a dependency on the migration files.
Everything else looks good to me.

promgen/migrations/0051_customlabel_customlabelinstance.py
hoangpn: All the migration files in this PR are generated by running `makemigration`. 👀 
Should I touch them?

vincent-olivert-riera: Yes, modify them and test them. There should not be any problem.
We use the `dependencies` list to determine on which order the migrations should be applied.

hoangpn: I got it. 🙇

hoangpn: @vincent-olivert-riera 
I have made the change, tested and did a force-push. Please continue reviewing me PR. 🙇

vincent-olivert-riera (approved):

PR #778, opened 2026-09-15
Fix bug that activateTabFromHash() causes promgen.js to stop working

hoangpn:
In the promgen.js file, we have a function named activateTabFromHash() used to activate a Bootstrap tab based on the hash value of the URL. This function is also called once when the page loads to display the tab based on the current hash value.

However, this behavior causes errors for pages with an unrecognized expression hash value in the URL, such as the API Specification page. As a result, the JavaScript code fails to run afterward.

To resolve this issue, we escaped any character that has a special meaning in a CSS selector to ensure that the hash can be used as a valid selector.


promgen/static/js/promgen.js
```diff
@@ -137,34 +137,38 @@ function initSelect2() {
...
+      document.querySelectorAll("form").forEach(form => {
+        form.action = form.action.replace(/#.*$/, "") + hash;
+      });
+      document.querySelectorAll("a.btn:not([data-toggle])").forEach(btn => {
+        btn.href = btn.href.replace(/#.*$/, "") + hash;
+      });
 
+    }
+  } catch (error) {
+    console.error("Failed to activate tab from hash", error);
```
vincent-olivert-riera: If this function is only meant to run when the page contains Bootstrap tabs, instead of putting everything inside a try/catch block, shouldn't instead check if the page actually has Bootstrap tabs, and if it doesn't, then return without doing anything.

Throwing errors on pages where this function is not meant to run doesn't look good, I think.

```
function activateTabFromHash() {
  // if the page does not have Bootstrap tabs
  bye bye

  // rest of the code here
}
```

vincent-olivert-riera (changes requested):

promgen/static/js/promgen.js
hoangpn: I got it. Let me implement in this direction.
Thank you so much for your suggestion! 🙇

hoangpn: I implemented the check. I realized that the original function still works fine if the Bootstrap tabs is missing.
The error happened because the hash contains some special characters. 🙇 Therefore, I implemented the necessary code and edited the commit message.

promgen/static/js/promgen.js
```diff
@@ -137,8 +137,15 @@ function initSelect2() {
 // Activate a tab based on the hash value of the URL
 // For example, if the URL is `http://example.com#tab2`, it will activate the tab with id `tab2`.
 function activateTabFromHash() {
-  const hash = window.location.hash;
-  if (hash && $(hash).length) {
+  var hash = window.location.hash;
```
vincent-olivert-riera:
```suggestion
  let hash = window.location.hash;
```

vincent-olivert-riera (changes requested):

vincent-olivert-riera (approved):

PR #779, opened 2026-09-16
Support users to select a specific expiration for the API Token

hoangpn:
AS-IS:
image
image

TO-BE:
image
image


vincent-olivert-riera:
I think the user will select the expiration date thinking on their current time zone.
For instance, if I select `2026-09-17 09:30`, I will expect that date to be based for my time zone.
Therefore, when we display the dates in the UI, we should display them also using the user's local time, not UTC. I don't want the user to think and calculate the real expiry date.

hoangpn:
> I think the user will select the expiration date thinking on their current time zone. For instance, if I select `2026-09-17 09:30`, I will expect that date to be based for my time zone. Therefore, when we display the dates in the UI, we should display them also using the user's local time, not UTC. I don't want the user to think and calculate the real expiry date.

Thank you so much for your suggestion. I have added the necessary change. 🙇

vincent-olivert-riera (approved):

PR #782, opened 2026-09-21
Add validations when updating project

hoangpn:
We have made some changes to improve validations when updating important fields of an existing Project: the `owner` and the `parent Service`.


vincent-olivert-riera (approved):
> On the other hand, users with Editor permission and above can easily change the parent Service of a Project. [...]

Well spotted 👍

